PR #3, opened 2018-10-07
Add subscription support to inpage provider via provider-engine

danfinlay:
Related to https://github.com/MetaMask/metamask-extension/pull/5450


kumavis:
- Doesn't look like events are being re-emitted from the provider-engine to the inpage-provider (easy to fix)
- Not sure about the side effects of re-adding the provider-engine's polling at this level (uncertain)

PR #4, opened 2018-11-20
Add events for account and network change

danfinlay:
This was embarrassingly easy for how long we didn't do it.

### DIFF
--- a/index.js
+++ b/index.js
@@ -15,6 +15,8 @@ util.inherits(MetamaskInpageProvider, SafeEventEmitter)
 
 function MetamaskInpageProvider (connectionStream) {
   const self = this
+  self.selectedAddress = undefined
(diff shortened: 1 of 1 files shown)


danfinlay:
There's an issue:

On account change, the store is updated twice:
- Once with an undefined account
- Once with the new account

This would probably be unexpected behavior if you just wanted to know that a change was happening, so I'd like to fix that.

That said, this is still probably better than what every dev does today to detect account changes.

danfinlay:
Note that currently MetaMask reloads web3-touching pages on network change, but we've also [said we were going to stop doing that](https://medium.com/metamask/breaking-change-no-longer-reloading-pages-on-network-change-4a3e1fd2f5e7), and rolling in a network change event is a good precursor to that breaking change.

PR #11, opened 2019-11-14
Make compatible with EIP 1193, EIP 1102, and EIP 2255 (i.e. LoginPerSite)

rekmarks:
This makes the inpage provider compatible with EIP 1193, EIP 1102, and EIP 2255 (i.e., [this branch of the extension](https://github.com/MetaMask/metamask-extension/tree/LoginPerSite)). It also sets us up for the December 16 API changes (basically, delete everything marked with `TODO:deprecate:2019-12-16`).

Some noteworthy changes:
- use `yarn`
- metadata handling moved from extension to here
- `ethereum._metamask` proxy moved from extension to here
- `ethereum.send` fully implemented per EIP 1193
- major version bump to `4.0.0`
  - changes are breaking if the background does not use EIP 2255

We probably shouldn't merge this and publish the package until the extension's `LoginPerSite` PR is approved, but we definitely review it now.


index.js
```diff
@@ -1,174 +1,551 @@
...
-      self.selectedAddress = state.selectedAddress
-      self.emit('accountsChanged', [self.selectedAddress])
+    if ('isUnlocked' in state && state.isUnlocked !== _state.isUnlocked) {
+      _state.isUnlocked = state.isUnlocked
+    }
+
+    // Emit chainChanged event on chain change
+    if ('chainId' in state && state.chainId !== this.chainId) {
+      this.chainId = state.chainId
+      this.emit('chainChanged', this.chainId)
```
danfinlay, line 93: Does this work with our current network controller? Last time I was trying to do this, it needed some refactoring.

index.js
```diff
@@ -1,174 +1,551 @@
...
+  jsonRpcConnection.events.on('notification', payload => {
+    if (payload.method === 'wallet_accountsChanged') {
+      this._handleAccountsChanged(payload.result)
+    } else {
+      this.emit('data', null, payload)
+    }
   })
 
-  // Work around for https://github.com/metamask/metamask-extension/issues/5459
-  // drizzle accidently breaking the `this` reference
```
danfinlay, line 93: Does drizzle still work after this change? If not, have we notified them of it?

index.js
```diff
@@ -1,174 +1,551 @@
...
+ * @param {Array<any>} [params] - If given a method name, the method's parameters.
+ * @returns {Promise<any>} - A promise resolving to the result of the method call.
+ */
+MetamaskInpageProvider.prototype.send = function (methodOrPayload, params) {
+
+  // construct payload object
+  let payload
+  if (params !== undefined) {
+
+    // wrap params in array out of kindness
```
danfinlay, line 232: Pro tip acquired

index.js
```diff
@@ -1,174 +1,551 @@
...
-    result: result,
+    result,
+  }, 'result')
+}
+
+/**
+ * Internal method for calling EIP-1102 eth_requestAccounts.
+ * Attempts to call eth_accounts before requesting the permission.
+ */
+MetamaskInpageProvider.prototype._requestAccounts = function () {
```
danfinlay: Maybe could have been a little cleaner as an async method.

Also why using internal sendAsync instead of regular send here?

index.js
```diff
@@ -1,174 +1,551 @@
...
+
+    if (!payload.jsonrpc) {
+      payload.jsonrpc = '2.0'
+    }
+
+    if (
+      payload.method === 'eth_signTypedData' &&
+      !_state.sentWarnings.signTypedData
+    ) {
+      console.warn(messages.warnings.signTypedDataDeprecation)
```
danfinlay: Maybe this is just me, but if we aren’t planning to remove it or change how it works, why clutter the logs with it? A few warnings here I think we could just drop, or maybe reduce to info.

index.js
```diff
@@ -1,174 +1,551 @@
...
-function logStreamDisconnectWarning (remoteLabel, err) {
-  let warningMsg = `MetamaskInpageProvider - lost connection to ${remoteLabel}`
-  if (err) warningMsg += '\n' + err.stack
-  console.warn(warningMsg)
-  const listeners = this.listenerCount('error')
-  if (listeners > 0) {
-    this.emit('error', warningMsg)
+  // TODO:deprecate:2019-12-16
+  // handle web3
+  if (window.web3) {
```
danfinlay: Might want to also check for window.web3.eth, for ultimate safety.

danfinlay (approved):
Looks good! A few things you might want to tighten up or discuss, but I don’t see anything worth blocking. Very clean, pleasant to read!

index.js
rekmarks: Yes, I confirmed that chainId is on the inpage provider, and that the event works (although it only has the wrong name in production).

index.js
rekmarks: I assume so, since we closed that issue (https://github.com/metamask/metamask-extension/issues/5459) after Truffle/Drizzle said it should be fine.

index.js
rekmarks: Done

index.js
rekmarks: `_sendAsync` does some preflight checks/preprocessing before handing requests off to `rpcEngine.handle`.

I ended up moving this method to an async middleware in the extension background, where it should live.

index.js
rekmarks: Upon review, I removed the warning for that method completely. The rest are either being removed or things that I hope to remove.

Gudahtt:
It looks like this needs to be rebased; there are conflicts.

rekmarks:
@Gudahtt fixed!

.nvmrc
```diff
@@ -0,0 +1 @@
+v10.16.0
```
Gudahtt: Nit: We _just_ updated to v10.17.0 in the `metamask-extension` repo. Maybe worth using here as well.

index.js
```diff
@@ -1,184 +1,491 @@
...
+
+// private state, kept here in part for use in the _metamask proxy
+const _state = {
+
+  sentWarnings: {
+    enable: false,
+    experimentalMethods: false,
+    isConnected: false,
+    sendAsync: false,
+    // TODO:deprecate:2019-12-16
```
Gudahtt: I guess this date could be updated 👀

index.js
```diff
@@ -1,184 +1,491 @@
...
+    : resolve(response.result)
+}
 
 module.exports = MetamaskInpageProvider
 
-util.inherits(MetamaskInpageProvider, SafeEventEmitter)
+inherits(MetamaskInpageProvider, SafeEventEmitter)
+
+// private state, kept here in part for use in the _metamask proxy
+const _state = {
```
Gudahtt: Could this state be moved onto the `MetamaskInpageProvider` instance? The fact that it's in module scope effectively makes `MetamaskInpageProvider` into a singleton.

index.js
```diff
@@ -1,184 +1,491 @@
...
+      },
+
+      /**
+       * Deprecated. Will be removed on 2019-12-16.
+       * Asynchronously determines if this domain is currently enabled
+       *
+       * @returns {Promise<boolean>} - Promise resolving to true if this domain is currently enabled
+       */
+      isApproved: async () => {
+        return Array.isArray(_state.accounts) && _state.accounts.length > 0
```
Gudahtt: Huh. Nothing async going on here, as it's not waiting for the first state update from the background anymore.

Technically this is a breaking change, as this will now return false negatives just like `isEnabled` does. I'm not sure how much I _care_ that it's a breaking change, as it's experimental and due for removal, but... this does kinda make this method useless.

index.js
```diff
@@ -1,184 +1,491 @@
...
+  return new Proxy(
+    {
+
+      /**
+       * Determines if MetaMask is unlocked by the user.
+       *
+       * @returns {Promise<boolean>} - Promise resolving to true if MetaMask is currently unlocked
+       */
+      isUnlocked: async () => {
+        return _state.isUnlocked
```
Gudahtt: Same case as for `isApproved` - I think this'll fail if called early enough, in the same manner as `isEnabled`

index.js
Gudahtt: We could probably wait for the first call of `_handleAccountsChanged`, and then the behavior would be the same 🤔

index.js
```diff
@@ -1,184 +1,491 @@
...
+  else if (window.web3 && typeof window.web3.eth === 'object') {
+    window.web3.eth.defaultAccount = this.selectedAddress
   }
 }
 
-function noop () {}
+/**
+ * Gets experimental _metamask API as Proxy.
+ */
+function getExperimentalApi (instance) {
```
Gudahtt, line 437: I've added a `registerOnboarding` experimental method in the `one-click-onboarding` branch that'll need to be ported over here I suppose 🤔 

I'll consider this further - maybe we can stash that somewhere else. It feels a little inappropriate to go here.

Gudahtt (commented):
Here's the first part of my review - I'm still only part-way through.

index.js
rekmarks: Yeah, I don't think it's a big deal either, but we shouldn't break it. It now waits for the first emission of `accountsChanged`.

index.js
rekmarks: Now waits for first `_publicConfigStore` update.

index.js
rekmarks: Hm, yeah that's more of a conundrum. Would we need to set up the `onboardingStream` in the inpage provider for that to work? If not, I think it is ultimately fine if it lives here.

On the other hand, there is perhaps a case to be made for experimental methods closely tied to the background to be defined in the background in `inpage.js`? Although, in the end, everything still "lives" in the same place at runtime?

index.js
```diff
@@ -1,184 +1,507 @@
...
```
Gudahtt, line 437: This should probably be a method I guess? Otherwise the `this` references will refer to this function, not the provider instance.

index.js
```diff
@@ -1,184 +1,507 @@
...
-    // Emit accountsChanged event on account change
-    if ('selectedAddress' in state && state.selectedAddress !== self.selectedAddress) {
-      self.selectedAddress = state.selectedAddress
-      const accounts = state.selectedAddress ?
-        [state.selectedAddress] :
-        []
-      self.emit('accountsChanged', accounts)
+  this._publicConfigStore = new ObservableStore({ storageKey: 'MetaMask-Config' })
+
+  // chainChanged and networkChanged events
```
Gudahtt: Nit: this function also handles reacting to the wallet locking/unlocking

index.js
```diff
@@ -1,184 +1,507 @@
...
+      method: methodOrPayload,
+      params,
+    }
   } else {
-    return self._sendSync(payload)
+
+    if (typeof methodOrPayload === 'string') {
+      payload = {
+        method: methodOrPayload,
+        params,
```
Gudahtt: Nit: `params` can be omitted here, since it's undefined.

index.js
```diff
@@ -1,184 +1,507 @@
...
   }
+
+  if (!Array.isArray(payload)) {
+
+    if (!payload.jsonrpc) {
+      payload.jsonrpc = '2.0'
+    }
+
+    if (
+      payload.method === 'eth_accounts' ||
```
Gudahtt, line 359: It seems a bit weird to me that a read-only method would have a side-effect 🤔 

Are there existing cases where the accounts can change without the provider knowing?

index.js
```diff
@@ -1,184 +1,507 @@
...
   )
 
   // subscribe to metamask public config (one-way)
-  self.publicConfigStore = new ObservableStore({ storageKey: 'MetaMask-Config' })
-
-  // Emit events for some state changes
-  self.publicConfigStore.subscribe(function (state) {
-
-    // Emit accountsChanged event on account change
-    if ('selectedAddress' in state && state.selectedAddress !== self.selectedAddress) {
```
Gudahtt, line 41: Hmm. So changes to `selectedAddress` are no longer transmitted via this stream. How is this now being communicated to the provider?

index.js
```diff
@@ -1,184 +1,508 @@
...
-  self.networkVersion = undefined
-  self.chainId = undefined
 
   // super constructor
-  SafeEventEmitter.call(self)
+  SafeEventEmitter.call(this)
+
+  // private state, kept here in part for use in the _metamask proxy
+  this._state = {
+      sentWarnings: {
```
Gudahtt: Nit: indentation is off here

index.js
```diff
@@ -1,184 +1,508 @@
...
     }
   })
 
   pump(
     mux.createStream('publicConfig'),
-    asStream(self.publicConfigStore),
-    logStreamDisconnectWarning.bind(this, 'MetaMask PublicConfigStore')
+    asStream(this._publicConfigStore),
+    // RPC requests should still work if only this stream fails
+    logStreamDisconnectWarning.bind(this, 'MetaMask PublicConfigStore'),
```
Gudahtt, line 108: Interesting 🤔 This does severely hamper the abilities of the provider though - it can't fulfill it's duties anymore in sending events like `chainChanged`.

Maybe this is fine. I'm not sure that emitting `close` would be better. It does seem unsatisfying though.

Ah well - it matches the previous behavior at least, so I have no objections.

index.js
```diff
@@ -1,184 +1,508 @@
...
   )
 
   // ignore phishing warning message (handled elsewhere)
   mux.ignoreStream('phishing')
 
+  // setup own event listeners
+
+  // EIP-1193 subscriptions
```
Gudahtt: I don't see a reference to a `data` event in EIP-1193 - perhaps I'm missing something?

index.js
```diff
@@ -1,184 +1,508 @@
...
+ * @returns {Promise<any>} - A promise resolving to the result of the method call.
+ */
+MetamaskInpageProvider.prototype.send = function (methodOrPayload, params) {
+
+  // construct payload object
+  let payload
+  if (params !== undefined) {
+
+    // wrap params in array out of kindness
+    if (!Array.isArray(params)) {
```
Gudahtt: This seems to be a breaking change. Previously `.send`'s second parameter was a callback. Anybody now passing a callback to `.send` will find that it's wrapped in an array and treated as a set of parameters.

index.js
```diff
@@ -1,184 +1,508 @@
...
+      params = [params]
+    }
+
+    // method must be a string if params were supplied
+    // we will throw further down if it isn't
```
Gudahtt: Similarly, this will break any dapps passing a request object and a callback to `.send`

Gudahtt (changes requested):

index.js
rekmarks: Ah, I missed updating the `_state` references to `instance._state` after making `_state` into an property of `MetamaskInpageProvider`. IIRC, I was having problems with accessing `this` inside the `Proxy`, but in any event, it's a helper method that doesn't need to be bound to the prototype.

Let me know if you feel strongly about it.

index.js
rekmarks: Due to `LoginPerSite`, the preference controller's `selectedAddress` is no longer of relevance to external domains. Whatever accounts we get from the permissions controller via `eth_accounts` and JSON RPC notifications are what we care about.

index.js
Gudahtt: Hey whatever works is fine by me. That was just a suggestion for how to fix it, I hadn't thought about the implications of it being within a proxy.

index.js
rekmarks: Hm, I'm not following. Those lines - which relate to the `publicConfig` stream - work exactly the same now as they did before. In testing, `chainChanged` and `networkChanged` are emitted as expected. Are you referring to something else?

Gudahtt: I mean that the `publicConfig` stream is being treated here as a non-critical, even though it'd mean that the provider would only be half-working if it failed.

Gudahtt: To be more specific: `_handleDisconnect` isn't being called here in the event of failure.

I'm not totally sure how this worked before 🤔. I see that there's a similar method called `_onClose` in `createStandardProvider`, but I don't see it being called anywhere...

index.js
rekmarks: Hm, strange. This was originally there, and we basically emitted a `data` event that we created a listener for which then emitted a `notification` event. I changed this to just emit a `notification` event.

index.js
rekmarks: Good catch, fixing that.

index.js
rekmarks: I'm pondering that. I may just have done it as a defensive measure. TBD...

index.js
Gudahtt: Right, of course, sorry I misrepresented myself there. What I should have asked is: how is the provider notified when the user switches accounts?

Also, the act of switching accounts was how the provider used to be notified of _new_ accounts being created, because upon creating a new account you'd also switch to it. Is the provider still notified when a new account has been created?

This relates somewhat to this comment: https://github.com/MetaMask/metamask-inpage-provider/pull/11#discussion_r349292382

I don't think the old provider handled this particularly well either, as I'm pretty sure you could remove accounts without the provider being notified until you happened to switch accounts as well.

index.js
rekmarks: Thinking about this one, I don't know if there's a case where a single stream fails; I think they all go or none of them do, and I've never seen any evidence to the contrary. That said, my reasoning here was that, even if the `publicConfig`-related functionality breaks, you can still send RPC requests. If we interpret `connect` and `close` as related to Ethereum network connectivity, should we continue to not emit `close`?

Gudahtt: Ah, right. I guess `isConnected` and the `close` event are more about signalling that the provider is connected, and less about signalling that the provider itself is fully working. That's a good point. The way you've done it here is probably best then.

src/utils.js
```diff
@@ -0,0 +1,96 @@
...
+  // don't do anything to Promises
+  if (obj instanceof Promise) return obj
+
+  const defineOpts = {
+    configurable: true, writable: true, enumerable: false,
+  }
+
+  // strange wrapping of Promise functions to fully emulate .then behavior,
+  // specifically Promise chaining
+  // there may be a simpler way of doing it, but this works
```
Gudahtt, line 73: Interesting solution! This is missing a `.finally` though.

That could be added in the same way, but... I was wondering, did you try creating a real Promise and attaching the `obj` properties to it instead? Sorta like the reverse of what you've done here. That should even hold up to `result instanceof Promise` checks, which would be nice.

index.js
```diff
@@ -1,184 +1,508 @@
...
+
+  if (!this._state.sentSiteMetadata) {
+    sendSiteMetadata(this._rpcEngine)
+    this._state.sentSiteMetadata = true
   }
+
+  if (!Array.isArray(payload)) {
+
+    if (!payload.jsonrpc) {
+      payload.jsonrpc = '2.0'
```
Gudahtt, line 355: Cool! This will probably fix this issue: https://github.com/MetaMask/metamask-extension/issues/7433

src/utils.js
rekmarks: Good catch on the `.finally`.

I think you're highlighting the fundamental reason why we can't really have our cake and eat it too. If we can imagine someone doing `result instanceof Promise` and expecting it to be `true` - in the new usage of the method - we can also imagine someone expecting it to be `false` - in the old usage of the method.

I'm not sure what's best here.

index.js
rekmarks: We'll have to review this in light of changes to `LoginPerSite`, but I don't think the inpage provider will ever be out of sync with the extension background. In other words, it'll call `_handleAccountsChanged`, but it ought to never do anything.

Nevertheless, if we ever break something such that it becomes out of sync, is it worth keeping as a defensive measure? It's a read-only method, sure, but we are in an async environment.

index.js
```diff
@@ -1,184 +1,508 @@
...
+/**
+ * Internal RPC method. Forwards requests to background via the RPC engine.
+ * Also remap ids inbound and outbound.
+ */
+MetamaskInpageProvider.prototype._sendAsync = function (payload, userCallback) {
+
+  let cb = userCallback
+
+  if (!this._state.sentSiteMetadata) {
+    sendSiteMetadata(this._rpcEngine)
```
Gudahtt: This seems like an odd time to send site metadata - especially since it's called in a fire-and-forget way here, so the initial `requestAccounts` call might go through and trigger the connect notification before the metadata has arrived.

Would it be possible to move this to the constructor instead, after establishing a connection with MetaMask?

rekmarks: Yeah, moved it to the constructor.

rekmarks:
@Gudahtt, just pushed stuff, mostly related to things I marked resolved. Only significant changes worth looking over ought to be the `send` method. I kind of hate it, but it will remain something of a mess until we've stopped supporting the `send(requestObject, callback)` signature.

src/utils.js
Gudahtt: Fair point, but I had thought that by taking this "thenable" approach, we had already disregarded cases like that.

After all, even adding a `.then` will break a dapp that _does_ call `.then` but expects it to fail :smile:  But that's a fairly unreasonable expectation to have of this method, so I'm OK with breaking it.

Is there a case to be made for a consumer of the provider relying upon `result instanceof Promise` being `false`? I would hope that anybody checking whether the result is a Promise is then proceeding to handle the Promise (otherwise why bother check?). This might break dapps that _think_ they're supporting Promises correctly but aren't, but... I'm not sure that's on us.

Gudahtt: Oh, also, we'll also ideally want to make sure `result.toString()` and `JSON.stringify(result)` return the same thing they did before too. Hmm.

Maybe your way is easier :sweat_smile:  This is just temporary after all - whatever's easiest to get working with `.then`, `.catch`, `.finally`, `toString`, and `JSON.stringify` is probably our best choice.

index.js
Gudahtt: OK,  I'll take a look at the `LoginPerSite` branch later to look for these cases. It should be communicated by `wallet_accountsChanged` notifications I think?

If we ever _do_ end up relying upon this side-effect, it means something is already broken. Generally speaking I'd rather things blow up in production and us find out quickly, than them be subtly wrong in a way that takes longer for us to discover and fix. I don't feel that strongly about it though - feel free to leave it in as a backup if you want, I'm not going to block on it.

index.js
```diff
@@ -1,184 +1,508 @@
...
-// handle sendAsync requests via asyncProvider
-// also remap ids inbound and outbound
-MetamaskInpageProvider.prototype.sendAsync = function (payload, cb) {
-  const self = this
+/**
+ * Sends an RPC request to MetaMask. Resolves to the result of the method call.
+ * May reject with an error that must be caught by the caller.
+ *
+ * @param {(string|Object)} methodOrPayload - The method name, or the RPC request object.
+ * @param {Array<any>} [params] - If given a method name, the method's parameters.
```
Gudahtt, line 194: Nit: This should be updated to reflect that it can also be a Function - at least for now

index.js
```diff
@@ -1,184 +1,508 @@
...
+    // handle send(object, callback), an alias for sendAsync(object, callback)
+    if (typeof params === 'function') {
+      return this._sendAsync(methodOrPayload, params)
+    }
+
+    payload = methodOrPayload
+
+    // TODO:deprecate:2020-01-13
+    // backwards compatibility: "synchronous" methods
+    if ([
```
Gudahtt: I think `!params` should be added to this condition? Synchronous send was only accessible if the second param was falsy

src/siteMetadata.js
```diff
@@ -0,0 +1,91 @@
...
+  const metaTitle = document.querySelector('head > meta[name="title"]')
+  if (metaTitle) {
+    return metaTitle.content
+  }
+
+  if (document.title && document.title.length > 0) {
+    return document.title
+  }
+
+  return window.location.hostname
```
Gudahtt, line 58: Cool, this seems like a good fallback.

src/siteMetadata.js
```diff
@@ -0,0 +1,91 @@
...
+  }
+
+  return null
+}
+
+/**
+ * Returns whether the given resource exists
+ * @param {string} url the url of the resource
+ */
+function resourceExists (url) {
```
Gudahtt, line 87: This is a great idea! 🙌

index.js
rekmarks: Good point; I too am a fan of fail-hard. How about this, inside `_handleAccountsChanged`?
```javascript
if (isEthAccounts) {
  log.error(
    'MetaMask: Accounts may be out of sync. Please report this bug.',
    accounts,
  )
}
```

src/utils.js
rekmarks: Right! Good point about `toString` and `JSON.stringify`. I think Finlay may have raised that as a reason originally. We'll keep the current hack and hope for the best!

src/utils.js
```diff
@@ -0,0 +1,98 @@
...
+ *
+ * @returns {Function} json-rpc-engine middleware function
+ */
+function createErrorMiddleware () {
+  return (req, res, next) => {
+    next(done => {
+      const { error } = res
+      if (!error) {
+        return done()
+      // legacy eth_accounts behavior
```
Gudahtt: Interesting - where was this legacy behavior? I had trouble finding something that matched this.

test/utils.js
```diff
@@ -0,0 +1,75 @@
...
+    id: 2512950,
+  }
+
+  const customThenable = obj => makeThenable(obj, 'result')
+  const promiseThenable = async obj => obj.result
+
+  const customRes = await runThenableTests(customThenable)
+  const promiseRes = await runThenableTests(promiseThenable)
+
+  Object.entries(customRes).forEach(async ([k, v1]) => {
```
Gudahtt: I suspect that this might not complete before `t.end()` is called? As nothing is `await`-ing each of these loop operations.

You could update this to use an async iterator, or `await` in a for loop, or map the entries to `Promise.all` - one of those.

test/utils.js
```diff
@@ -0,0 +1,75 @@
...
+      t.ok(v2 instanceof Promise, 'promiseThenable direct return is a Promise')
+      const v2res = await v2
+      t.ok(v2res === target, 'promiseThenable direct return resolves to target value')
+    } else {
+      if (v1 instanceof Promise) {
+        t.ok(v2 instanceof Promise, 'value1 instanceof Promise -> value2 instanceof Promise')
+        const r1 = await v1
+        const r2 = await v2
+        t.deepEqual(r1, r2, 'promises resolve to the same values')
+      // } else if (v1 === target) {
```
Gudahtt: Nit: leftover comment

index.js
Gudahtt: Sure, that's a great idea! It could be challenging to detect these cases without a warning like that.

Though I take it that `isEthAccounts` would instead be a comparison between the results and `this._state.accounts`, or something along those lines.

rekmarks: Cool! Yeah, I'm keep it in inside another if-block. I think we can put this one to bed!

index.js
rekmarks: I'm not sure if it was documented originally, and I don't want to encourage anyone by indicating otherwise.

index.js
Gudahtt: It wasn't really even meant to be an "experimental" method in the first place - I just put it there because it was convenient 😬 

After giving this a bit more thought, I think I can accomplish what I set out to do with that `registerOnboarding` method in a simpler manner by adding a custom RPC method (e.g. `wallet_registerOnboarding`). I'll talk it over with some people tomorrow, but either way I think we should get these changes in first, and merge `one-click-onboarding` second.

src/utils.js
rekmarks: Good catch. For a while on `LoginPerSite`, we were throwing EIP 1193 `unauthorized` errors in the background if the permission was not granted. We're always returning an array now, so I'm removing that clause.

index.js
Gudahtt: It's still documented: https://metamask.github.io/metamask-docs/API_Reference/Ethereum_Provider#ethereum.send(options%2C-callback)-(to-be-replaced)

test/utils.js
rekmarks: Hah, it was working before, but lord only knows why. I did a `.map` inside a `Promise.all` as I've been liking that pattern recently 💅

index.js
rekmarks: Ah.

Let's still ignore it? 😅

index.js
rekmarks: Woah, I don't think we're cleaning up permissions when accounts are removed. If we do that, a notification should be produced.

Regarding switching and creating new accounts though, the account will only change from the dapp's perspective if the permission changes. I'm not sure exactly what we're doing with those in the UI, but I'll make sure to discuss it with @danjm.

Gudahtt (approved):
Phew 😌 , I think that's everything! This looks good to me. Both of the problems we encountered at the hackathon are fixed now (emitting `accountsChanged` upon unlock and accepting a callback to `.send`).

index.js
Gudahtt: Ah right, that's a good point. The site would need to explicitly be given permission to see the new account. We have that bulk assign permission screen, but it'd need to be revisited for each new account a user wanted to give access to.

Well, that makes it even simpler then, hopefully. I'll keep an eye out for this while looking over the extension `LoginPerSite` branch.

index.js
Gudahtt: Oh second thought, I'm going to merge `one-click-onboarding` since it's already gone through so many rounds of review and been approved. I also realized that it won't be possible to use an RPC message, as I need to grab the tab ID from the extension message API...

I'll just take care of the conflicts myself by putting up a separate PR to `LoginPerSite` (either this repo or the extension or both).

rekmarks: Thanks for letting me know. If that needs any integration with the new UI you should probably hold off on the PR until `LoginPerSite-ui` is merged into `LoginPerSite`. Just FYI!

PR #12, opened 2019-11-26
Send site metadata on dom content loaded to ensure metadata availability

danjm:
This PR fixes an issue with the retrieval of site metadata. Prior to this PR, it seems that `document.title` is undefined at the time that `sendSiteMetadata` is called. By only calling `sendSiteMetaData` when the dom content is ready (on a `DOMContentLoaded`), we ensure that (at least) `document.title` will be defined if available.

### DIFF
--- a/index.js
+++ b/index.js
@@ -20,7 +20,7 @@ const {
 } = require('./src/utils')
 
 // resolve response.result, reject errors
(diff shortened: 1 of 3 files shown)


Gudahtt (approved):
Looks good to me!

Gudahtt (approved):
Seems reasonable!

PR #13, opened 2019-12-14
Let eth_accounts sync accounts on unlock

rekmarks:
We've noticed the "accounts out of sync" error thrown frequently. A likely majority of these cases were due to the provider using `eth_accounts` to sync accounts when the extension becomes unlocked or the provider becomes initialized. This PR adds a boolean flag to allow `eth_accounts` to sync accounts without warning in those two cases.

Also removes a recently added, non-standard, and undocumented event we were emitting on extension unlock.

### DIFF
--- a/index.js
+++ b/index.js
@@ -77,7 +77,6 @@ function MetamaskInpageProvider (connectionStream) {
 
     if ('isUnlocked' in state && state.isUnlocked !== this._state.isUnlocked) {
       this._state.isUnlocked = state.isUnlocked
-      this.emit('wallet_isUnlocked', this._state.isUnlocked)
       if (!this._state.isUnlocked) {
         // accounts are never exposed when the extension is locked
         this._handleAccountsChanged([])
@@ -87,6 +86,7 @@ function MetamaskInpageProvider (connectionStream) {
           this._sendAsync(
             { method: 'eth_accounts', params: [] },
             () => {},
+            true // indicating that eth_accounts _should_ update accounts
           )
         } catch (_) {}
       }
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -426,9 +431,9 @@ MetamaskInpageProvider.prototype._handleAccountsChanged = function (accounts, is
 
     // we should always have the correct accounts even before eth_accounts
     // returns, except if the method is called before we're fully initialized
-    if (isEthAccounts && this._state.accounts !== undefined) {
+    if (isEthAccounts && !isInternal && this._state.accounts !== undefined) {
       log.warn(
```
rekmarks: Should this be an error?

Gudahtt (commented):
This would certainly be an improvement for now, but I'd still prefer that we block on getting the normal `accountsChanged` event from the background, rather than sending an _additional_ message requesting for a second copy of these accounts.

rekmarks:
@Gudahtt Me too. The problem is that there is that the inpage provider does not receive an `accountsChanged` notification from the background when the extension becomes unlocked.

The fix for that is to remove the public config store and only use JSON RPC notifications to communicate extension state to the inpage provider. It's on my TODOs, albeit pretty far down.

Gudahtt:
Ohhh I see, I forgot about that. OK, well this makes sense in the meantime I suppose. We can remove it after the public config store is removed.

index.js
Gudahtt: I think so, yes.

Gudahtt (approved):
LGTM! Though I do agree that that message should probably be promoted to `log.error` instead of `log.warn`

PR #14, opened 2020-02-04
More validation in _sendAsync; optional metadata sending

rekmarks:
### Main

This PR typechecks `payload.method` in the inpage provider's error middleware. I originally did this in `_sendAsync`, throwing an error on failure. To prevent unexpected errors to be thrown in consumer code, I moved the typecheck to the middleware so that the error appears on the RPC response object, as it would have previously.

I also decided against typechecking or setting `payload.params` in `_sendAsync`. `params` are an array in Ethereum JSON RPC by convention only; JSON RPC 2.0 allows `params` to be an array, object, or omitted. In other words, the receiving end should handle it. I implemented a temporary fix in Mobile for now, and will remove `web3-provider-engine` later.

All remaining thrown errors in this module either cause Promise rejections or have always been there.

### Misc

These changes also allow the consumer of this module to specify whether `wallet_sendDomainMetadata` should be sent, by means of a boolean argument to the constructor. This will be useful for both plugins and Mobile, for instance.

### DIFF
--- a/index.js
+++ b/index.js
@@ -32,7 +32,7 @@ module.exports = MetamaskInpageProvider
 
 inherits(MetamaskInpageProvider, SafeEventEmitter)
 
-function MetamaskInpageProvider (connectionStream) {
+function MetamaskInpageProvider (connectionStream, shouldSendMetadata = true) {
 
   // super constructor
   SafeEventEmitter.call(this)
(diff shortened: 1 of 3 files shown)


Gudahtt (approved):
LGTM!

index.js
```diff
@@ -251,12 +255,12 @@ MetamaskInpageProvider.prototype.send = function (methodOrPayload, params) {
...
-  // typecheck payload and payload.method
+  // typecheck payload, payload.method, and payload.params
   if (
-    Array.isArray(payload) ||
-    typeof params === 'function' ||
     typeof payload !== 'object' ||
-    typeof payload.method !== 'string'
+    Array.isArray(payload) ||
+    !Array.isArray(params) ||
+    !payload.method || typeof payload.method !== 'string'
```
Gudahtt: Is this being checked twice now (here and in the error middleware)? :thinking: 
If so, perhaps it shouldn't be checked here? The error in the middleware seems more appropriate.

rekmarks: Good point, the other conditions can stay for now.

Gudahtt (approved):
LGTM!

PR #17, opened 2020-02-07
Convert inpage provider to Class

rekmarks:
Convers the inpage provider to a class, and binds its instance methods. Should exhibit no behavioral changes except now work even when `web3.js` tries to ruin it for no reason.

**Misc**
- Minor version bump to `4.1.0`
- Bump `eth-json-rpc-errors`
- Use `@metamask/eslint-config`


danfinlay:
If I’m not mistaken, a similar situation is the historic reason for the proxy we wrap around the provider in the extension (truffle was doing something funny, deleting a property):
https://github.com/MetaMask/metamask-extension/blob/8e307f8d714bfd06051f68add4ef03af7eb8226b/app/scripts/inpage.js#L66-L70
This Proxy allows us to prevent bad consumers from manipulating the provider in unexpected ways. We probably could just @agoric/harden the provider before exporting, on the inpage-provider module itself, and solve both these problems of consumer-mutation.
I don’t see which part of this PR does that change (the class wrapping as described), though, so I’m not sure.

rekmarks:
@danfinlay 

The problem was that `enable` (at least, who knows what else) became unbound.

I've confirmed that the chosen solution works, and it is to bind the functions, see here: https://github.com/MetaMask/metamask-inpage-provider/blob/51f3d4a2da7f22d8e628f920470b22045a6b278d/index.js#L61-L68

I'm aware of the Proxy wrapping we already do and the reason therefore. I'd definitely be down to harden it, although I'm hesitant because of all the voodoo web3 does. I wouldn't be surprised if we break it by doing that.

danfinlay (approved):

Gudahtt (approved):
LGTM!

PR #18, opened 2020-02-08
Fix incorrect eth_accounts update-accounts-error

### LINKED ISSUE #16: MetaMask: 'eth_accounts' unexpectedly updated accounts
Reporting as requested.

I first noticed the issue after merging develop in https://github.com/MetaMask/metamask-extension/pull/7987#issuecomment-583123624 so I suspect the issue is caused in this patch https://github.com/MetaMask/metamask-extension/pull/7997 .

I reproduce the issue on develop by signing in to metamask, visiting https://ratifier.trusttoken.com and adding that site to "Connected Sites". It shows in the console for the page.

rekmarks:
Fixes #16 

See #16 for more details.

Version bump to `4.1.1`

### DIFF
--- a/index.js
+++ b/index.js
@@ -439,7 +439,7 @@ module.exports = class MetamaskInpageProvider extends SafeEventEmitter {
 
       // we should always have the correct accounts even before eth_accounts
       // returns, except in cases where isInternal is true
-      if (isEthAccounts && !isInternal) {
+      if (isEthAccounts && this._state.accounts !== undefined && !isInternal) {
         log.error(
           `MetaMask: 'eth_accounts' unexpectedly updated accounts. Please report this bug.`,
           accounts,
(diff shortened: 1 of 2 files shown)


Gudahtt (approved):
LGTM!

This is a pretty janky solution, but [it's a known problem that has been discussed before](https://github.com/MetaMask/metamask-inpage-provider/pull/13#pullrequestreview-346019949).

dotnetspec:
I still get this error when I log in/out of MM:
Linux Mint 18.3
MetaMask Version
8.0.5
Google Chrome
Version 84.0.4147.89 (Official Build) (64-bit)
I'm unable to upgrade to 8.0.6 as the Chrome Webstore only makes 8.0.5 available.

Gudahtt:
@dotnetspec Interesting - which dapp did you see this on? I am unable to reproduce this on Chrome with v8.0.5 but the dapps I'm testing with might be doing something differently.

PR #20, opened 2020-02-28
call _handleAccountsChange when selectedAddress changed in config

nugaon:
When I used the Metamask Inpage Provider in my chrome extension, it didn't get any wallet_accountsChanged event. Also it can be useful if the selectedAddress changes in ConfigStore, can be handled correctly by calling the  _handleAccountsChange method.

### DIFF
--- a/index.js
+++ b/index.js
@@ -106,6 +106,11 @@ module.exports = class MetamaskInpageProvider extends SafeEventEmitter {
         this.emit('chainIdChanged', this.chainId) // TODO:deprecate:2020-Q1
       }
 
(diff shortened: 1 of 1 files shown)


rekmarks:
Hi @nugaon, the reason we don't check for `selectedAddress` in the `publicConfigStore` state is because we've removed it in MetaMask. In fact, we are going to remove the `publicConfigStore` entirely, and replace it completely with JSON RPC notifications.

Consequently, we will not modify the existing `publicConfigStore` functionality in this module, and I am closing this PR.

PR #29, opened 2020-04-05
Roll back to production API; fixup

rekmarks:
This PR rolls back to the Provider API currently in production.

- Un-deprecate `sendAsync`
- Deprecate `send`; restore its API to what's currently in production
  - What's currently in production is an ungodly mix between the inpage provider and the extension's `StandardProvider`
  - Specifically, per the `package.json` and `yarn.lock` files of [the latest release](https://github.com/MetaMask/metamask-extension/releases/tag/v7.7.8):
    - [`metamask-inpage-provider@3.0.0`](https://github.com/MetaMask/metamask-inpage-provider/blob/156f283f8e1c00526033fcfc3659c8e679d2b756/index.js#L98-L107)
    - [`createStandardProvider.js`](https://github.com/MetaMask/metamask-extension/blob/master/app/scripts/createStandardProvider.js#L37-L73)
  - Note that `StandardProvider` is removed from `develop`; [the present `inpage.js`](https://github.com/MetaMask/metamask-extension/blob/develop/app/scripts/inpage.js#L101) just wraps the inpage provider in a Proxy
- Removes `makeThenable` and related tests
  - Disable unit tests for CI, because those were all the tests we had
  - Adding unit tests for `index.js`is still a TODO
- Minor code reorganization and fixup


.eslintrc.js
```diff
@@ -22,7 +22,9 @@ module.exports = {
   parserOptions: {
     ecmaVersion: 2018,
   },
-  rules: {},
+  rules: {
+    'require-await': 'off',
```
Gudahtt, line 26: :+1: 
I've opened a PR [here](https://github.com/MetaMask/eslint-config/pull/37) to turn it off in the main repo, but this works for now.

index.js
```diff
@@ -206,92 +195,47 @@ module.exports = class MetamaskInpageProvider extends SafeEventEmitter {
...
+            { method: methodOrPayload, params: callbackOrArgs },
+            getRpcPromiseCallback(resolve, reject),
+          )
+        } catch (error) {
+          reject(error)
+        }
       })
+    } else if (
+      typeof methodOrPayload === 'object' &&
+      typeof callbackOrArgs === 'function'
```
Gudahtt, line 229: This seems more strict than what we have in prod. If I'm reading this correctly, prod would call `sendAsync` if the callback was truthy - regardless what the first param was or whether the second is a function.

I'm not sure if it _actually works_ in those cases though :thinking:

Gudahtt: It looks like if `methodOrPayload` was something other than an object, or if `callbackOrArgs` was something truthy other than a function, it would have blown up in `json-rpc-engine`. So being pickier here about the types seems like it shouldn't break anything.

Gudahtt (approved):
LGTM!

PR #30, opened 2020-04-13
Implement new 1193 spec, fixup

rekmarks:
Implements [the new EIP 1193 specification](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1193.md).

- Adds
  - Methods
    - `request(args)`, the only future-proof way of making RPC requests
  - Events
    - `message`, a replacement for `notification`
    - `disconnect`, a replacement for `close`
- Deprecates (in addition to existing deprecations, with warning messages)
  - Events
    - `chainIdChanged`
    - `close`
    - `networkChanged`
    - `notification`
- Changes
  - Experimental methods:
    - `sendBatch` -> `requestBatch`

### DIFF
--- a/src/MetamaskInpageProvider.js
+++ b/src/MetamaskInpageProvider.js
@@ -220,6 +220,8 @@ module.exports = class MetamaskInpageProvider extends SafeEventEmitter {
   //====================
 
   /**
+   * Experimental. The signature of this method may change without warning, pending EIP 1193.
+   *
    * Submits an RPC request to MetaMask for the given method, with the given params.
    * Resolves with the result of the method call, or rejects on error.
    *
--- a/src/MetamaskInpageProvider.js
+++ b/src/MetamaskInpageProvider.js
@@ -269,8 +269,16 @@ module.exports = class MetamaskInpageProvider extends SafeEventEmitter {
   /**
    * We override the following event methods so that we can warn consumers
    * about deprecated events:
-   *   on, once
+   *   addListener, on, once, prependListener, prependOnceListener
    */
+
+  /**
+   * @inheritdoc
+   */
+  addListener (eventName, listener) {
+    this._warnOfDeprecation(eventName)
+    return super.addListener(eventName, listener)
+  }
 
   /**
    * @inheritdoc
@@ -286,6 +294,22 @@ module.exports = class MetamaskInpageProvider extends SafeEventEmitter {
   once (eventName, listener) {
     this._warnOfDeprecation(eventName)
     return super.once(eventName, listener)
+  }
+
+  /**
+   * @inheritdoc
+   */
+  prependListener (eventName, listener) {
+    this._warnOfDeprecation(eventName)
+    return super.prependListener(eventName, listener)
+  }
+
+  /**
+   * @inheritdoc
+   */
+  prependOnceListener (eventName, listener) {
+    this._warnOfDeprecation(eventName)
+    return super.prependOnceListener(eventName, listener)
   }
 
   //====================
--- a/src/MetamaskInpageProvider.js
+++ b/src/MetamaskInpageProvider.js
@@ -53,7 +53,6 @@ module.exports = class MetamaskInpageProvider extends SafeEventEmitter {
         experimentalMethods: false,
         isConnected: false,
         send: false,
-        // sendAsync: false,
         // events
         events: {
           chainIdChanged: false,
@@ -255,6 +254,16 @@ module.exports = class MetamaskInpageProvider extends SafeEventEmitter {
         getRpcPromiseCallback(resolve, reject),
       )
     })
+  }
+
+  /**
+   * Submit a JSON-RPC request object and a callback to make an RPC method call.
+   *
+   * @param {Object} payload - The RPC request object.
+   * @param {Function} callback - The callback function.
+   */
+  sendAsync (payload, cb) {
+    this._rpcRequest(payload, cb)
   }
 
   /**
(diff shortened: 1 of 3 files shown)


Gudahtt (dismissed):
Could we hold off on deprecating `sendAsync` until after `request` is stabilized? We should probably mark `request` as "experimental" until the EIP is finalized as well.

rekmarks:
@Gudahtt I undeprecated `sendAsync`, but I'd rather not mark `request` experimental, as it would only discourage its use, and I believe it is very unlikely to change. How we communicate about this new API in docs and announcements is a different story.

Gudahtt:
As long as we should communicate in some way that we might break `request` without warning. Otherwise we'd be in a very uncomfortable position if the spec did change, however unlikely.

Gudahtt:
As long as we should communicate in some way that we might break `request` without warning. Otherwise we'd be in a very uncomfortable position if the spec did change, however unlikely.

Gudahtt:
As long as we should communicate in some way that we might break `request` without warning. Otherwise we'd be in a very uncomfortable position if the spec did change, however unlikely.

src/MetamaskInpageProvider.js
```diff
@@ -45,16 +45,25 @@ module.exports = class MetamaskInpageProvider extends SafeEventEmitter {
...
-    // private state, kept here in part for use in the _metamask proxy
+    // private state
     this._state = {
       sentWarnings: {
+        // methods
         enable: false,
         experimentalMethods: false,
         isConnected: false,
         send: false,
+        // sendAsync: false,
```
Gudahtt: Nit: this comment can be removed

src/MetamaskInpageProvider.js
```diff
@@ -488,16 +564,26 @@ module.exports = class MetamaskInpageProvider extends SafeEventEmitter {
...
   /**
-   * TODO:deprecation:remove
+   * Submit a JSON-RPC request object and a callback to make an RPC method call.
+   *
+   * @param {Object} payload - The RPC request object.
+   * @param {Function} callback - The callback function.
+   */
+  sendAsync (payload, cb) {
+
+    // if (!this._state.sentWarnings.sendAsync) {
```
Gudahtt: Nit: these comments can be removed

src/MetamaskInpageProvider.js
```diff
@@ -488,16 +564,26 @@ module.exports = class MetamaskInpageProvider extends SafeEventEmitter {
...
+    this._rpcRequest(payload, cb)
+  }
+
+  /**
+   * DEPRECATED
    * Internal backwards compatibility method.
    */
   _legacySend (payload) {
 
-    if (!this._state.sentWarnings.sendSync) {
```
Gudahtt, line 496: Why is this being removed?

rekmarks: Because we're already warning in `send`, which is the only place where this is called, for a particular combination of parameters. It used to be named `_sendSync`, perhaps that's better?

Gudahtt: Ah I see - gotcha. So this was accidentally missed in #29

rekmarks: Ah yeah, it would've caused a double warning. Anyway, I reverted the name to `_sendSync`, and addressed your nits.

src/MetamaskInpageProvider.js
```diff
@@ -342,12 +433,10 @@ module.exports = class MetamaskInpageProvider extends SafeEventEmitter {
         },
 
         /**
-         * Make a batch request.
+         * Make a batch RPC request.
          */
-        // eslint-disable-next-line require-await
-        sendBatch: async (requests) => {
+        requestBatch: async (requests) => {
```
Gudahtt, line 462: So `sendBatch` is being renamed as part of this as well? That's definitely worth calling out in the final commit message / PR description, and also ensuring it's in the changelog.

It is experimental, but still we don't want to break that without documenting it somewhere.

src/MetamaskInpageProvider.js
```diff
@@ -214,6 +266,28 @@ module.exports = class MetamaskInpageProvider extends SafeEventEmitter {
...
+   */
+  on (eventName, listener) {
+    this._warnOfDeprecation(eventName)
+    return super.on(eventName, listener)
+  }
+
+  /**
+   * @inheritdoc
+   */
+  once (eventName, listener) {
```
Gudahtt, line 294: There are other `EventEmitter` methods for attaching event listeners (e.g. `prependListener`, `prependOnceListener`, `addListener`)

src/messages.js
```diff
@@ -1,16 +1,24 @@
...
+    autoReloadDeprecation: `MetaMask: MetaMask will soon stop reloading pages on network change.\nFor more information, see: https://medium.com/metamask/no-longer-reloading-pages-on-network-change-fbf041942b44 \nSet 'ethereum.autoRefreshOnNetworkChange' to 'false' to silence this warning: https://metamask.github.io/metamask-docs/API_Reference/Ethereum_Provider#ethereum.autorefreshonnetworkchange`,
+    // deprecated methods
+    enableDeprecation: `MetaMask: 'ethereum.enable()' is deprecated and will be removed in the future. Please use "ethereum.request({ method: 'eth_requestAccounts' })" instead.\nFor more information, see: https://eips.ethereum.org/EIPS/eip-1102`,
+    isConnectedDeprecation: `MetaMask: 'ethereum.isConnected()' is deprecated and will be removed in the future. Please listen for the relevant events instead.\nFor more information, see: https://eips.ethereum.org/EIPS/eip-1193`,
+    sendDeprecation: `MetaMask: 'ethereum.send(...)' is deprecated and will be removed in the future. Please use 'ethereum.request(method: string, params: Array<any> | Object)' instead.\nFor more information, see: https://eips.ethereum.org/EIPS/eip-1193`,
```
Gudahtt: We should probably be recommending `sendAsync` here, not `request`. Or at least `sendAsync(...) or `request(...)`

src/messages.js
```diff
@@ -1,16 +1,24 @@
...
+    sendAsyncDeprecation: `MetaMask: 'ethereum.sendAsync(...)' is deprecated and will be removed in the future. Please use 'ethereum.request(method: string, params: Array<any> | Object)' instead.\nFor more information, see: https://eips.ethereum.org/EIPS/eip-1193`,
```
Gudahtt: Nit: this can be removed, since we aren't using it yet.

src/MetamaskInpageProvider.js
rekmarks: I think it's an exhaustive list with the ones you mentioned. Adding.

src/MetamaskInpageProvider.js
rekmarks: `sendBatch` never made it into production, but I noted it in the PR description nevertheless.

Gudahtt: Ah, sorry, I had wrongly assumed it was in prod. That's not so bad then.

Gudahtt (approved):
LGTM!

PR #31, opened 2020-04-14
Add global initialization event

rekmarks:
- Adds a global initialization event for asynchronous injection: `ethereum#initialized`
- Restructures the repo and the module's exports
  - Adds exported initialization utility methods
  - Compare work with [the extension's `inpage.js`](https://github.com/MetaMask/metamask-extension/blob/develop/app/scripts/inpage.js)

### DIFF
--- a/src/MetamaskInpageProvider.js
+++ b/src/MetamaskInpageProvider.js
@@ -43,7 +43,7 @@ module.exports = class MetamaskInpageProvider extends SafeEventEmitter {
 
     this.isMetaMask = true
 
-    this.setMaxEventListeners(maxEventListeners)
+    this.setMaxListeners(maxEventListeners)
 
     // private state, kept here in part for use in the _metamask proxy
     this._state = {
--- a/src/MetamaskInpageProvider.js
+++ b/src/MetamaskInpageProvider.js
@@ -22,11 +22,28 @@ const {
 
 module.exports = class MetamaskInpageProvider extends SafeEventEmitter {
 
-  constructor (connectionStream, shouldSendMetadata = true) {
+  /**
+   * @param {Object} connectionStream - A Node.js stream
+   * @param {Object} opts - An options bag
+   * @param {number} opts.maxEventListeners - The maximum number of event listeners
+   * @param {boolean} opts.shouldSendMetadata - Whether the provider should send page metadata
+   */
+  constructor (
+    connectionStream,
+    { shouldSendMetadata = true, maxEventListeners = 100 },
+  ) {
+
+    if (
+      typeof shouldSendMetadata !== 'boolean' || typeof maxEventListeners !== 'number'
+    ) {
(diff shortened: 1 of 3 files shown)


Gudahtt (changes requested):
Could you move the version bump to a separate PR?

rekmarks:
@Gudahtt done

Gudahtt (dismissed):

src/MetamaskInpageProvider.js
```diff
@@ -22,12 +22,29 @@ const {
...
+      typeof shouldSendMetadata !== 'boolean' || typeof maxEventListeners !== 'number'
+    ) {
+      throw new Error('Invalid options.')
+    }
 
     super()
 
     this.isMetaMask = true
 
+    this.setMaxEventListeners(maxEventListeners)
```
Gudahtt: I think this function is misnamed (https://nodejs.org/docs/latest-v10.x/api/events.html#events_emitter_setmaxlisteners_n)

```suggestion
    this.setMaxListeners(maxEventListeners)
```

src/initProvider.js
```diff
@@ -0,0 +1,55 @@
...
+  if (!connectionStream) {
+    throw new Error('Must provide a connection stream.')
+  }
+
+  let provider = new MetamaskInpageProvider(
+    connectionStream, { shouldSendMetadata, maxEventListeners },
+  )
+
+  if (proxyHandler) {
+    provider = new Proxy(provider, proxyHandler)
```
Gudahtt: Does this make sense as part of the provider API? 🤔 All this does is wrap it in a Proxy, which the caller could do more easily.

I'd suggest either removing it, or replace this with a flag to wrap it in a Proxy that prevents property deletion (e.g. the thing we use the Proxy for).

Gudahtt (changes requested):

src/initProvider.js
rekmarks: I added a default proxy handler that's the same as the one in the extension's `inpage.js`.

The benefit of handling the Proxy here is that we can update it in one place and use it in both mobile and the extension.

Gudahtt: This API still seems.... very strange? As a consumer of this package, why would I want to pass in a `proxyHandler`? What utility would it have? What would the consequences be of passing in a `proxyHandler` (i.e. it'd potentially break compatibility with older version of `web3`, but that's not easy to tell from looking at this API).

A flag called `web3Compatibility` or `preventPropertyDeletion` or something like that might make it more clear what's going on (defaulting to `true`). Then the consumer could disable it and handle the proxy wrapping by themselves, or they could leave it enabled and create their own proxy... whichever. It would at least be clearer then _what_ we're doing and _why_.

rekmarks: I did `preventPropertyDeletion`

Gudahtt (approved):
LGTM!

PR #33, opened 2020-04-17
Reorganize repo for future changes

rekmarks:
- Move main class to own file in `src` folder
  - A perfect copy, except for the updated local module paths
- Preserve `index.js` exports


Gudahtt (approved):
LGTM! Thanks for separating this - this'll make for a much more forgiving diff 😌

PR #37, opened 2020-04-22
Rename to @metamask/inpage-provider

rekmarks:
- Rename to `@metamask/inpage-provider`
  - Fix description in `package.json`


.nvmrc
```diff
@@ -1 +1 @@
-v10.17.0
```
Gudahtt: Could this be moved to a separate PR? It doesn't seem related to the rename

Gudahtt (changes requested):
The README has a few references to the old name as well

.nvmrc
rekmarks: Please no. I updated the PR title and description such that it is related

Gudahtt (approved):
LGTM!

.nvmrc
rekmarks: For the record: I submitted

PR #38, opened 2020-04-22
Loosen .nvmrc

rekmarks:
Loosen `.nvmrc` to `v10` per other repos


Gudahtt (dismissed):
LGTM!

I don't think we use any core Node.js libraries in this repo, so I doubt this would fail on earlier versions of v10

Gudahtt (approved):

PR #39, opened 2020-05-07
Prevent some property overwrites in proxy

rekmarks:
This prevents some internal state properties and the non-deprecated parts of our API from being overwritten, if the Provider is proxy-wrapped.

It won't solve existing overwrites of e.g. `ethereum.send`, but we don't care about that method anyway, and in this way, we won't break existing dapps.

### DIFF
--- a/src/initProvider.js
+++ b/src/initProvider.js
@@ -18,7 +18,7 @@ function initProvider ({
   protectProperties = true,
   shouldSendMetadata = true,
   shouldSetOnWindow = true,
-}) {
+} = {}) {
 
   const PROTECTED_PROPERTIES = new Set([
     '_handleAccountsChanged',
--- a/src/initProvider.js
+++ b/src/initProvider.js
@@ -20,15 +20,9 @@ function initProvider ({
   shouldSetOnWindow = true,
 } = {}) {
 
+  // public, non-deprecated properties
   const PROTECTED_PROPERTIES = new Set([
-    '_handleAccountsChanged',
-    '_handleDisconnect',
     '_metamask',
-    '_publicConfigStore',
-    '_rpcEngine',
-    '_rpcRequest',
-    '_sendSync',
-    '_state',
     'isMetaMask',
     'request',
   ])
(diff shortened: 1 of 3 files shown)


brad-decker (dismissed):
👍 LGTM

src/initProvider.js
```diff
@@ -6,18 +6,32 @@ const MetamaskInpageProvider = require('./MetamaskInpageProvider')
...
   connectionStream,
   maxEventListeners = 100,
-  preventPropertyDeletion = true,
+  protectProperties = true,
   shouldSendMetadata = true,
   shouldSetOnWindow = true,
-}) {
+} = {}) {
+
+  const PROTECTED_PROPERTIES = new Set([
```
Gudahtt, line 24: Should this contain all private methods? I noticed that `_warnOfDeprecation` is missing. So are methods we're inheriting from `SafeEventEmitter`.

Maybe it'd be easier to list the non-protected properties 🤔

rekmarks: I'm not sure what's best here. I tried to do that initially, but ended up breaking some `EventEmitter` functionality due to a private property (IIRC, `_eventCount` or something like that) being purposefully overwritten by the `EventEmitter` implementation. While we could fix that particular problem, it made me worried about inadvertently breaking stuff. What do you think about the risks here?

Gudahtt: Oh right - we might want a lot of these things to be mutable internally.

Gudahtt: What's bizarre is that this is thing we have to care about in the first place 😵. 

At some point we have to trust the user of the provider to not mess with internals, or risk it breaking. Maybe we could go in the other direction then, and only protect the methods likely to be set (e.g. `isMetaMask` and `request`, the public ones). It seems fairly burdensome to maintain a complete list of all internal private properties and methods, just to provide a friendlier development environment for the users I'm least interested in helping frankly.

Gudahtt (dismissed):
This seems OK to me, but I think it'd be easier on us to only protect public methods/properties

Gudahtt (dismissed):
LGTM!

Gudahtt (approved):

PR #41, opened 2020-05-20
Fix send method return value

rekmarks:
In the production extension, `ethereum.send(method: string, params?: Array<any>)` returns the entire JSON RPC `response` object, not `response.result`.

We want to keep this to avoid breaking `send`, so this has now been fixed.

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -34,12 +34,12 @@ function createErrorMiddleware () {
   }
 }
 
-// resolve response.result, reject errors
-const getRpcPromiseCallback = (resolve, reject) => (error, response) => {
+// resolve response.result or response, reject errors
+const getRpcPromiseCallback = (resolve, reject, resolveResponse = false) => (error, response) => {
   if (error || response.error) {
     reject(error || response.error)
   } else {
-    Array.isArray(response)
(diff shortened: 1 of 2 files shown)


src/utils.js
```diff
@@ -34,12 +34,12 @@ function createErrorMiddleware () {
   }
 }
 
-// resolve response.result, reject errors
-const getRpcPromiseCallback = (resolve, reject) => (error, response) => {
+// resolve response.result or response, reject errors
+const getRpcPromiseCallback = (resolve, reject, resolveResponse = false) => (error, response) => {
```
Gudahtt: `unwrapResult` might be a better name for this? The promise is getting resolved either way here.

Gudahtt (dismissed):

src/utils.js
Gudahtt: Well, technically nothing is getting resolved here. The promise is resolved elsewhere, and the response is returned here, unwrapped or not.

rekmarks: Can't argue with that! 83b4220

Gudahtt (approved):
LGTM!

PR #44, opened 2020-05-21
Add default connection stream

rekmarks:
- Add default connection stream
- `initProvider` -> `initializeProvider`

### DIFF
--- a/src/initializeProvider.js
+++ b/src/initializeProvider.js
@@ -1,18 +1,32 @@
+const PostMessageDuplexStream = require('post-message-stream')
 const MetamaskInpageProvider = require('./MetamaskInpageProvider')
+
+module.exports = {
+  initializeProvider,
+  setGlobalProvider,
+}
+
+const getDefaultConnectionStream = () => {
+  return new PostMessageDuplexStream({
+    name: 'metamask-inpage',
+    target: 'metamask-contentscript',
+  })
+}
 
 /**
    * Initializes a MetamaskInpageProvider and (optionally) sets it on window.ethereum.
    *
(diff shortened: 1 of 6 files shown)


Gudahtt:
Why include a default stream? I thought asking it to be passed in was an intentional design choice - to keep the provider completely removed from the transport.

rekmarks:
@Gudahtt a default is included for consumers where our injection fails, such as the `script-src: self` CSP in Firefox.

A consumer can still pass in a connection stream if they want, of course.

Gudahtt:
Fair enough; I suppose it's not reasonable to expect consumers to know exactly how to construct a `postMessage` stream for the extension.

src/initializeProvider.js
```diff
@@ -1,18 +1,32 @@
...
 
+module.exports = {
+  initializeProvider,
+  setGlobalProvider,
+}
+
+const getDefaultConnectionStream = () => {
+  return new PostMessageDuplexStream({
+    name: 'metamask-inpage',
+    target: 'metamask-contentscript',
```
Gudahtt, line 12: This won't work yet - not until the stream is renamed in the extension too, right?

rekmarks: Yeah, that follows a suggestion by @danfinlay to rename them to something less generic.

I have a draft PR on the extension (MetaMask/metamask-extension#8640) that renames the extension and incorporates the changes in #43. (Of course, we can do a more staged merge on the extension side if we wish, but I reckon we should probably just do a single major version bump of this package.)

rekmarks:
We don't need this.

PR #47, opened 2020-05-22
Modify property overwrite protection

rekmarks:
Modifies the properties protected from overwrites by the following logic:
- Even though we want to, we shouldn't protect properties that existed before 5.0.0
  - Therefore, `isMetaMask` and `_metamask` are no longer protected
- Since we continue to protect `request`, let's also protect the two new private properties it relies on to function:
  - `_rpcRequest`
  - `_rpcEngine`


Gudahtt (approved):
LGTM!

PR #49, opened 2020-06-01
Add site origin to metadata

rekmarks:
Add `origin: window.location.origin` to site metadata object.


Gudahtt (changes requested):
I don't think this is required. The extension already gets the `origin` from the connection itself. More importantly, that `origin` is guaranteed to be correct by the platform, but this one could be spoofed.

rekmarks:
@Gudahtt, yeah you're right, it's not.

PR #51, opened 2020-06-02
Fix args.params type checks

rekmarks:
Fix typo in error message and fix type check of `args.params` (it was erroneously allowing `null`).


whymarrh (approved):

Gudahtt (approved):
LGTM!

PR #53, opened 2020-06-23
Remove protected properties

rekmarks:
If someone is sufficiently determined, they can break our provider. We'll continue to disallow deletions, but will have to live with assignments.


Gudahtt (approved):
LGTM!

PR #55, opened 2020-06-23
Add updated automatic reload on network change

rekmarks:
Per recent discussions, we decided to continue automatically reloading the page on network change. This PR introduces a new property (with a sensible name) for that purpose, and adds the automatic reload functionality to `initProvider`.

### Summary of Changes

- Add `reloadOnChainChange` property
- Use `Object.defineProperty` to define `reloadOnChainChange` and the (deprecated) `autoRefreshOnNetworkChange` alias on provider instances
- Update relevant messages

### DIFF
--- a/src/initProvider.js
+++ b/src/initProvider.js
@@ -18,7 +18,6 @@ function initProvider ({
 } = {}) {
 
   let wasAccessed = false
-  let isReloading = false
 
   if (!connectionStream) {
     throw new Error('Must provide a connection stream.')
@@ -30,10 +29,7 @@ function initProvider ({
 
   provider.once('chainChanged', () => {
     if (wasAccessed && provider.reloadOnChainChange) {
-      if (!isReloading) {
-        isReloading = true
-        setTimeout(() => window.location.reload(), 250)
-      }
+      window.location.reload()
     }
   })
 
--- a/src/initProvider.js
+++ b/src/initProvider.js
@@ -17,23 +17,47 @@ function initProvider ({
   shouldSetOnWindow = true,
 } = {}) {
 
+  let wasAccessed = false
+  let isReloading = false
+
   if (!connectionStream) {
     throw new Error('Must provide a connection stream.')
   }
 
-  let provider = new MetamaskInpageProvider(
+  const provider = new MetamaskInpageProvider(
     connectionStream, { shouldSendMetadata, maxEventListeners },
   )
 
-  provider = new Proxy(provider, {
-    deleteProperty: () => true, // some libraries, e.g. web3@1.x, mess with our API
+  provider.once('chainChanged', () => {
+    if (wasAccessed && provider.reloadOnChainChange) {
+      if (!isReloading) {
+        isReloading = true
+        setTimeout(() => window.location.reload(), 250)
+      }
+    }
+  })
+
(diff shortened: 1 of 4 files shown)


src/initProvider.js
```diff
@@ -17,23 +17,43 @@ function initProvider ({
...
-  let provider = new MetamaskInpageProvider(
+  const provider = new MetamaskInpageProvider(
     connectionStream, { shouldSendMetadata, maxEventListeners },
   )
 
-  provider = new Proxy(provider, {
-    deleteProperty: () => true, // some libraries, e.g. web3@1.x, mess with our API
+  provider.once('chainChanged', () => {
+    if (wasAccessed && provider.reloadOnChainChange) {
+      window.location.reload()
```
rekmarks: For the `web3` reload, we had a timeout that was activated if the `web3` object was "recently" activated (some fairly small number of milliseconds). I have no idea what good that's supposed to do, but I'm just noting it here for consideration.

src/utils.js
```diff
@@ -65,6 +66,47 @@ function logStreamDisconnectWarning (remoteLabel, err) {
...
+ *
+ * This helper exists because we want these properties to refer to an internal
+ * state value, via get/set handlers, and also be enumerable.
+ *
+ * @param {MetamaskInpageProvider} instance - The provider instance.
+ */
+function defineAutoReloadProperties (instance) {
+
+  const autoRefreshWarning = () => {
+    if (!instance._state.sentWarnings.autoRefresh) {
```
Gudahtt: Nit: it seems strange to me that this is a "utility" when it's operating on internal state of the provider. I'd generally assume that any instance state is mutated only in the module where the class is defined.

src/initProvider.js
Gudahtt: Huh. It would be nice to preserve the existing behavior exactly for the time being, but... that does sound a bit strange.

src/utils.js
rekmarks: Fair. Move it to the `MetamaskInpageProvider.js`?

Gudahtt: That would be much appreciated!

rekmarks: 0bcbc49

src/initProvider.js
rekmarks: Added `lastTimeUsed` per existing `web3` auto reload behavior in: 0bcbc49

Auto reloading is still set to `false` by default.

Gudahtt: >Auto reloading is still set to `false` by default.

Oh, huh, I missed this on my last review. Wouldn't this constitute a breaking change? I don't think we should change this right now - I would expect we'd change that at the same time as the `web3` removal (if ever).

rekmarks:
We've decided against making this change at the moment, largely due to `autoRefreshOnNetworkChange` only reloading the page if `window.web3` has been accessed.

That would mean that this change would be possibly breaking for existing dapps. In addition, we have plans for multi-network support that would require us to roll this back or modify the behavior.

rekmarks:
We've decided against making this change at the moment, largely due to `autoRefreshOnNetworkChange` only reloading the page if `window.web3` has been accessed.

That would mean that this change would be possibly breaking for existing dapps. In addition, we have plans for multi-network support that would require us to roll this back or modify the behavior.

PR #56, opened 2020-06-23
5.2.0

rekmarks:
Pending: #55

PR #57, opened 2020-06-24
Update autoRefresh property assignment and warnings

rekmarks:
### Changes

- Update the `autoRefreshOnNetworkChange` property warning
- Rename some variables
- Update some comments

Ref: #56

### DIFF
--- a/src/MetamaskInpageProvider.js
+++ b/src/MetamaskInpageProvider.js
@@ -62,7 +62,7 @@ module.exports = class MetamaskInpageProvider extends SafeEventEmitter {
         },
         // misc
         // TODO:deprecation:remove
-        autoReload: false,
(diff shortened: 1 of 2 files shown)


Gudahtt (commented):
Is this something we want to warn developers about right now? We don't have anything to suggest they do; it's not actionable. We also don't have a path to silence the warning, since we have a _different_ warning that suggests they should explicitly set this one.

rekmarks:
@Gudahtt, fair. Considering that it's not actionable for people who're already setting the property, and who may or may not be using `window.web3`, it's a bad idea. Changed the scope of this to merely update the warning message slightly, and change some comments.

Gudahtt (approved):
LGTM!

PR #59, opened 2020-06-29
Un-deprecate isConnected

rekmarks:
- Un-deprecate `ethereum.isConnected`
- Remove comment marking `request` as experimental
  - The review period of EIP-1193 has ended, and there will be no significant changes to its API

PR #60, opened 2020-06-29
5.2.1

rekmarks:
Ref: #59 

I think a patch bump is sufficient since all we're doing is removing a warning message?


Gudahtt (approved):
LGTM!

PR #62, opened 2020-07-04
Restore 'data' event

Gudahtt:
The `data` event was emitted from v3.0.0 of the inpage provider whenever a `notification` event was received from MetaMask. This event was removed in #11 because it was thought it was only used internally. Apparently `web3` was relying upon it though.

It has been restored exactly as it was in v3.0.0, except now a deprecation warning is shown in the console the first time a listener is added for this event.


rekmarks (approved):
Excellent!

PR #63, opened 2020-07-04
Restore notification event value

rekmarks:
Restores the emitted value of the `notification` event to its pre-`4.0.0` state, per: https://github.com/MetaMask/metamask-extension/blob/5e8a80e21d667b3c18a931e3da8f136b990bf14c/app/scripts/createStandardProvider.js#L32

Changing the value of the `notification` event was an unintended breaking change.

### DIFF
--- a/src/MetamaskInpageProvider.js
+++ b/src/MetamaskInpageProvider.js
@@ -184,7 +184,9 @@ module.exports = class MetamaskInpageProvider extends SafeEventEmitter {
         })
 
         // deprecated
-        this.emit('notification', (params && params.result) || payload)
+        if (params) {
+          this.emit('notification', params.result)
+        }
       }
     })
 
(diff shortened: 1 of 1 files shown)


src/MetamaskInpageProvider.js
```diff
@@ -169,15 +169,22 @@ module.exports = class MetamaskInpageProvider extends SafeEventEmitter {
...
+
         this.emit('message', {
-          type: payload.method,
-          data: payload.params,
+          type: method,
+          data: params,
         })
+
+        // deprecated
+        this.emit('notification', (params && params.result) || payload)
```
Gudahtt: Why the fallback to `payload`?

The old provider seemed to assume that `params.result` was always set for `eth_subscribe` messages. I'm not sure what we should do if it isn't, but emitting nothing seems like a safer choice than emitting the payload, which is a data structure that listeners might be totally unfamiliar with.
```suggestion
        this.emit('notification', params ? params.result : undefined)
```

Gudahtt (changes requested):

src/MetamaskInpageProvider.js
Gudahtt: Or maybe this is better

```suggestion
        if (params) {
          this.emit('notification', params.result)
        }
```

rekmarks: 96ba786

Gudahtt (approved):
LGTM!

PR #68, opened 2020-07-15
Add `publicConfigStore` alias

### LINKED ISSUE #67: `publicConfigStore` property renamed in v4.0.0
The provider property `publicConfigStore` was renamed in #11 and released in v4.0.0 of the inpage provider. We hadn't realized at the time that dapps were using this property directly, and that this change would break dapps.

We should restore the old property name for now as an alias, and warn that the property will soon be removed. Ultimately we'll still need to remove it at some point, since we're planning on replacing the `publicConfigStore` altogether.

Gudahtt:
An alias has been added to restore the `publicConfigStore` property that was renamed in #11. A warning is printed to the console upon first use, explaining that the property is deprecated and will be removed in the future.

Fixes #67

### DIFF
--- a/src/MetamaskInpageProvider.js
+++ b/src/MetamaskInpageProvider.js
@@ -63,6 +63,7 @@ module.exports = class MetamaskInpageProvider extends SafeEventEmitter {
         // misc
         // TODO:deprecation:remove
         autoRefresh: false,
+        publicConfigStore: false,
       },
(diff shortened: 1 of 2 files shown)


rekmarks (approved):
LGTM!

As a note, I believe that the `publicConfigStore` will be non-enumerable the way this is written, and I think that's fine.

PR #69, opened 2020-07-15
Add missing `data` event warning

### LINKED ISSUE #66: `data` event missing deprecation warning
The `data` event warns upon use because it is deprecated, but the deprecation message itself is missing. Instead `undefined` is printed to the console as a warning.

Gudahtt:
The `data` event was restored in #62 with a deprecation warning, but the deprecation warning message was missing. The message has now been added.

Fixes #66


rekmarks (approved):
LGTM!

PR #72, opened 2020-07-20
Emit accountsChanged after updating accounts state

rekmarks:
Ref: https://github.com/MetaMask/metamask-extension/issues/9027

We were emitting `accountsChanged` before all accounts-related state was updated. While it only affected deprecated state, we should still fix it.

This PR ensures that `accountsChanged` is only emitted _after_ all related state has been updated. In addition, it does some minor cleanup of the `_handleAccountsChanged` function.


whymarrh (approved):
LGTM!

Gudahtt (approved):
LGTM!

PR #73, opened 2020-07-21
Skip emitting `data` event for `wallet_accountsChanged`

Gudahtt:
The `data` event was being used by some dapps (via `web3@1.x` for example) in a way that we didn't anticpiate. Specifically these dapps were expecting the payload for this event to contain a `params` property, and they were breaking if this was absent.

We hadn't intended this event for public consumption in the first place, but we _especially_ didn't mean for `wallet_accountsChanged` to be sent to dapps. So for now the easiest solution is to not emit `data` for these events.

Currently the only other notification sent from the extension is `eth_subscription`, so the `data` event has been moved inside the block that checks for that type. The `params` check was removed for being redundant, since `eth_subscription` messages should always include a `params` property.

### DIFF
--- a/src/MetamaskInpageProvider.js
+++ b/src/MetamaskInpageProvider.js
@@ -171,13 +171,15 @@ module.exports = class MetamaskInpageProvider extends SafeEventEmitter {
     // json rpc notification listener
     jsonRpcConnection.events.on('notification', (payload) => {
 
-      this.emit('data', payload) // deprecated
-
       const { method, params, result } = payload
 
       if (method === 'wallet_accountsChanged') {
         this._handleAccountsChanged(result)
-      } else if (EMITTED_NOTIFICATIONS.includes(method)) {
+        return
+      }
+
+      if (EMITTED_NOTIFICATIONS.includes(method)) {
+        this.emit('data', payload) // deprecated
 
         this.emit('message', {
           type: method,
@@ -185,9 +187,7 @@ module.exports = class MetamaskInpageProvider extends SafeEventEmitter {
         })
 
         // deprecated
-        if (params) {
-          this.emit('notification', params.result)
-        }
+        this.emit('notification', params.result)
       }
     })
 


Gudahtt:
[Here is the line where `web3` makes this assumption](https://github.com/ethereum/web3.js/blob/1.x/packages/web3-core-requestmanager/src/index.js#L105) ([permalink](https://github.com/ethereum/web3.js/blob/b0b135cbb68faa3b66440b8e6829878fcc988ed2/packages/web3-core-requestmanager/src/index.js#L105)).

This was reported to us here: https://github.com/MetaMask/metamask-extension/issues/9013

src/MetamaskInpageProvider.js
```diff
@@ -171,14 +171,16 @@ module.exports = class MetamaskInpageProvider extends SafeEventEmitter {
...
-
       const { method, params, result } = payload
 
       if (method === 'wallet_accountsChanged') {
         this._handleAccountsChanged(result)
-      } else if (EMITTED_NOTIFICATIONS.includes(method)) {
+        return
+      }
+
+      this.emit('data', payload) // deprecated
```
rekmarks: If it's throwing on missing `params`, should we put this inside the `if` clause beginning on L190 below?

That would gate the emission by the `EMITTED_NOTIFICATIONS` enum too, but that should AFAIK be functionally identical to pre-v8 behavior.

Refs:
- https://github.com/MetaMask/metamask-extension/blob/develop/app/scripts/metamask-controller.js/#L1616
- https://github.com/MetaMask/eth-json-rpc-filters/blob/master/subscriptionManager.js/#L111-L120

Gudahtt: Hmm. I wasn't sure that that was the only existing case where `data` was emitted before. I can't find any other examples though.

Gudahtt: I'm inclined to leave it as-is for now, to avoid making any more assumptions about this event than were made in v3 of the inpage provider. I could probably be convinced to move it into the `if` condition though. Is there any reason that might be worth doing now?

rekmarks: I'm fine with that, but we should probably make sure that `params` are present on the `payload`?

Gudahtt: Right.... we do know that dapps using `web3@1.x` would break if we emitted a `data` payload without `params`. I don't love the idea of hard-coding this without knowing more about what payloads are possible though, as there could be similar assumptions made by consumers as well (e.g. some might presume that `params` is an object, some might assume the existence of other properties/sub-properties, etc.).

I'm coming around to your original suggestion now :thinking:. I'll do one last look around to see if there are any other cases aside from that. I found none in testing.

Gudahtt: I've reviewed every possible source of these events, and the two you linked are the only ones I can find. I'll make the update you originally suggested.

src/MetamaskInpageProvider.js
```diff
@@ -171,13 +171,15 @@ module.exports = class MetamaskInpageProvider extends SafeEventEmitter {
...
       const { method, params, result } = payload
 
       if (method === 'wallet_accountsChanged') {
         this._handleAccountsChanged(result)
-      } else if (EMITTED_NOTIFICATIONS.includes(method)) {
+        return
+      }
+
+      if (EMITTED_NOTIFICATIONS.includes(method)) {
+        this.emit('data', payload) // deprecated
```
rekmarks, line 182: Not inside the `if (params) { }` block?

Gudahtt: Oh - aren't `eth_subscription` notifications guaranteed to have a `params` property? I thought checking for that would have been redundant

rekmarks: Ah, they are, and that _is_ the only emitted notification at the moment.

How about we get rid of the `if (params)` clause entirely then for now, and bring `emit('message')` into the parent `if` clause?

Gudahtt: Ah right, sure that makes sense

rekmarks (approved):
👍

PR #74, opened 2020-07-21
Remove invalid test script

whymarrh:
This PR removes the invalid test script and `tape` devDep, as both aren't used.


rekmarks (approved):
👀 

image

PR #76, opened 2020-07-21
Handle multiple link elements with `rel=icon`

whymarrh:
Extracted from #75

It is very possible—even recommended—that there are multiple icons and if we're going to check that they actually exist, we should fallback to the other images listed.

e.g.:

```html
<link rel="icon" type="image/png" href="https://cdn.yourwebsite.com/favicon-16x16.png" sizes="16x16">
<link rel="icon" type="image/png" href="https://cdn.yourwebsite.com/favicon-32x32.png" sizes="32x32">
<link rel="icon" type="image/png" href="https://cdn.yourwebsite.com/favicon-96x96.png" sizes="96x96">
```

There is a subtle functionality change here in that the element matching `head > link[rel="shortcut icon"]` will no longer be checked first, but that isn't important, nor should it be. That said, not checking for `rel="shortcut icon"` first could also be an improvement since its usage isn't quite correct or modern:<sup>[\[1\]][1]</sup>

> For [historical reasons](https://www.w3.org/Bugs/Public/show_bug.cgi?id=12695), the HTML specification [now allows](https://html.spec.whatwg.org/multipage/semantics.html#rel-icon) the use of `shortcut` as a link relation if it’s immediately followed by a single U+0020 space character and the `icon` keyword.

  [1]:https://mathiasbynens.be/notes/rel-shortcut-icon

### DIFF
--- a/src/siteMetadata.js
+++ b/src/siteMetadata.js
@@ -68,17 +68,11 @@ function getSiteName (window) {
 async function getSiteIcon (window) {
   const { document } = window
 
-  // Use the site's favicon if it exists
-  let icon = document.querySelector('head > link[rel="shortcut icon"]')
-  if (icon && await resourceExists(icon.href)) {
-    return icon.href
-  }
-
-  // Search through available icons in no particular order
-  icon = Array.from(document.querySelectorAll('head > link[rel="icon"]'))
-    .find((_icon) => Boolean(_icon.href))
-  if (icon && await resourceExists(icon.href)) {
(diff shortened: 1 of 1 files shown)


rekmarks (dismissed):
LGTM!

whymarrh:
Fixed up the JSDoc

rekmarks (approved):

whymarrh:
Fixed up the JSDoc, again 😅

PR #77, opened 2020-07-21
Use mode:cors when checking for site icons

whymarrh:
This is a different approach than #75
Closes #78

This change updates the fetch mode we use when requesting site icons to use `cors` instead of `same-origin` as `same-origin` breaks when a site uses a different domain for their icons, e.g. `static.example.com` or `images.example.com`. While it is also possible that the 2nd origin here isn't correctly configured for CORS that's a fewer sites breaking.

### DIFF
--- a/src/siteMetadata.js
+++ b/src/siteMetadata.js
@@ -89,7 +89,7 @@ async function getSiteIcon (window) {
  * @param {string} url the url of the resource
  */
 function resourceExists (url) {
-  return fetch(url, { method: 'HEAD', mode: 'same-origin' })
+  return fetch(url, { method: 'HEAD', mode: 'cors' })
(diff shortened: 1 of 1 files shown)


whymarrh:
@rekmarks do you know why `same-origin` was used in the first place? It was introduced in #11.

whymarrh:
I'm also now questioning whether we need to check that the resources exist at all—what's the purpose of that? Presumable they can stop existing at any point, so this initial check maybe provides a false sense of certainty. [In the extension v6 we didn't do this extra check](https://github.com/MetaMask/metamask-extension/blob/cc1d506bd1082898e3dfd41e902bcfb4deb5503e/app/scripts/contentscript.js#L279-L298) for what it's worth.

PR #78, opened 2020-07-21
Use `img` elements for testing resource existence rather than `fetch`

whymarrh:
Closes #77

The previous implementation of `resourceExists` was using the `same-origin` mode for fetch but that breaks when a site uses another domain for their icons—`static.example.com`, `assets.example.com`, or `images.example.com`, for example.

There isn't a fetch mode that works here:<sup>[\[1\]][1]</sup>

- **`same-origin`** doesn't work for the reasons described
- **`cors`** there's no guarantee that the 2nd origin in question is correctly configured for CORS—a quick check on two sites confirmed this
- **`navigate`** isn't applicable here
- **`no-cors`** hides the `status` and `statusText` so we can't use those

With that, this PR loads the resources as images and tests whether that works. This will essentially be a `no-cors` request but with the status exposed through `onload`+`onerror`, which will suffice for our use case.

  [1]:https://fetch.spec.whatwg.org/#concept-request-mode

### DIFF
--- a/src/siteMetadata.js
+++ b/src/siteMetadata.js
@@ -70,14 +70,14 @@ async function getSiteIcon (window) {
 
   // Use the site's favicon if it exists
   let icon = document.querySelector('head > link[rel="shortcut icon"]')
-  if (icon && await resourceExists(icon.href)) {
+  if (icon && await imgExists(icon.href)) {
     return icon.href
   }
 
   // Search through available icons in no particular order
   icon = Array.from(document.querySelectorAll('head > link[rel="icon"]'))
     .find((_icon) => Boolean(_icon.href))
-  if (icon && await resourceExists(icon.href)) {
+  if (icon && await imgExists(icon.href)) {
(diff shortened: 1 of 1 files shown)


whymarrh:
This is the approach from #75, I think this is better than #77

rekmarks (changes requested):
I think this makes a lot of sense!

To follow up on questions in #77, I'm not positive why the `same-origin` restriction was added in the first place, likely out of an (over?)abundance of caution.

The whole point of `resourceExists` is to only send an `href` to the extension that actually points to something we can use. We have a similar fallback mechanism in the extension for when these links inevitably go stale, but I don't see any harm in doing some basic validation in the page so as to avoid sending complete garbage to the extension/mobile.

All of this said, let's rename `resourceExists` to `imageExists` in light of its new, more specific use?

rekmarks (approved):
`:guy_fieri_chef_kiss:`

PR #80, opened 2020-07-22
Validate constructor stream parameter

rekmarks:
For testing purposes, it's nice to have a specific error if an obviously invalid stream is passed. If it's not, things fail later in the constructor for unpredictable reasons.

- Adds `is-stream` as a dependency, which has no dependencies and is [tiny](https://bundlephobia.com/result?p=is-stream@2.0.0).
- Removes now-redundant stream argument check in `initProvider`

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,6 +28,7 @@
   "dependencies": {
     "eth-json-rpc-errors": "^2.0.2",
     "fast-deep-equal": "^2.0.1",
+    "is-stream": "^2.0.0",
     "json-rpc-engine": "^5.1.5",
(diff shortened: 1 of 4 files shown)


Gudahtt (approved):
LGTM

The inpage provider is where we should be the most averse to adding dependencies, since it gets added to `inpage.js` which blocks loading every single page. Adding any new JavaScript will increase the parse time even if it's not executed.

But... `is-stream` is indeed very tiny, basically negligible compared to the other dependencies we have already. So this seems fine to me. We can always inline it later if we really start trying to reduce the bundle size. If we do, there are a lot of other larger dependencies to remove first anyway.

PR #81, opened 2020-07-22
Use getters for all error messages

rekmarks:
~Pending: #80~

- Use getter functions for all error messages, for great testing convenience and consistency.
- Minor cleanup (formatting, include bad values with errors)

### DIFF
--- a/src/MetamaskInpageProvider.js
+++ b/src/MetamaskInpageProvider.js
@@ -24,7 +24,7 @@ const {
 module.exports = class MetamaskInpageProvider extends SafeEventEmitter {
 
   /**
-   * @param {Object} connectionStream - A Node.js stream
+   * @param {Object} connectionStream - A Node.js duplex stream
    * @param {Object} opts - An options bag
    * @param {number} opts.maxEventListeners - The maximum number of event listeners
    * @param {boolean} opts.shouldSendMetadata - Whether the provider should send page metadata
--- a/src/MetamaskInpageProvider.js
+++ b/src/MetamaskInpageProvider.js
@@ -9,6 +9,7 @@ const SafeEventEmitter = require('safe-event-emitter')
 const dequal = require('fast-deep-equal')
 const { ethErrors } = require('eth-json-rpc-errors')
 const log = require('loglevel')
+const { duplex: isDuplex } = require('is-stream')
 
 const messages = require('./messages')
 const { sendSiteMetadata } = require('./siteMetadata')
@@ -32,6 +33,10 @@ module.exports = class MetamaskInpageProvider extends SafeEventEmitter {
     connectionStream,
     { shouldSendMetadata = true, maxEventListeners = 100 } = {},
   ) {
+
+    if (!isDuplex(connectionStream)) {
+      throw new Error('Must provide a duplex stream.')
+    }
 
     if (
       typeof shouldSendMetadata !== 'boolean' || typeof maxEventListeners !== 'number'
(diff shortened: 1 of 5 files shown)


src/MetamaskInpageProvider.js
```diff
@@ -31,17 +31,20 @@ module.exports = class MetamaskInpageProvider extends SafeEventEmitter {
    */
   constructor (
     connectionStream,
-    { shouldSendMetadata = true, maxEventListeners = 100 } = {},
+    { maxEventListeners = 100, shouldSendMetadata = true } = {},
```
rekmarks, line 34: Reordered for great consistency with the English alphabet and the docstring.

src/MetamaskInpageProvider.js
```diff
@@ -31,17 +31,20 @@ module.exports = class MetamaskInpageProvider extends SafeEventEmitter {
...
+      throw new Error(messages.errors.invalidDuplexStream(), connectionStream)
     }
 
     if (
-      typeof shouldSendMetadata !== 'boolean' || typeof maxEventListeners !== 'number'
+      typeof shouldSendMetadata !== 'boolean' ||
+      typeof maxEventListeners !== 'number'
     ) {
-      throw new Error('Invalid options.')
+      throw new Error(messages.errors.invalidOptions(), {
```
Gudahtt: What syntax is this? The second parameter to the `Error` constructor is `filename` according to MDN.

rekmarks: You know when you invent an error syntax for an RPC error library and then confuse it with the native browser error syntax? 🤦

f14ba24

Gudahtt: It would be nice to pass in the parameters though, so that they're shown inline here, making it more difficult to forget to update them if the function signature changes

Gudahtt (dismissed):
LGTM!

Though I did have[ one more suggestion](https://github.com/MetaMask/inpage-provider/pull/81#discussion_r458960594) on how this could be improved

src/MetamaskInpageProvider.js
rekmarks: Hmm, like this?
```
    invalidOptions: (maxEventListeners, shouldSendMetadata) => `Invalid options. Received: { maxEventListeners: ${maxEventListeners}, shouldSendMetadata: ${shouldSendMetadata} }`,
```

Gudahtt: I guess that could work! I was thinking a stringified object.

rekmarks: I went with what I suggested, since if someone really messes up, the stringification of the object might throw.

Gudahtt: Oh I meant the interface specified as an object - not the params the user passed in. But sure this works too! At least it's a hint that it needs to be updated in two places

rekmarks: I thought that's what I did in f14ba24  / 28cd1d6 😛

Gudahtt: Just to clarify, I mean like this:

```
throw new Error(messages.errors.invalidOptions({ maxEventListeners: 'number', shouldSendMetadata: 'boolean' }))
```

Then the signature displayed in the message is directly here in the function, so future updates are all in the same general spot.

Edit: or like this:

```
throw new Error(messages.errors.invalidOptions('{ maxEventListeners: "number", shouldSendMetadata: "boolean" }'))
```

Gudahtt (approved):

PR #82, opened 2020-07-28
eth-rpc-errors@2.1.1

rekmarks:
- Removes `eth-json-rpc-errors`, which is deprecated
- Adds `eth-rpc-errors@2.1.1`
- Removes call to `serializeError` that has done nothing for quite some time
  - `serializeError` returns a serialized error, while it was used as though it modified the error in place.
- Minor cleanup of comments and order of logical operations in `createMiddleware`
  - All non-functional

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -1,13 +1,12 @@
 const EventEmitter = require('events')
 const log = require('loglevel')
-const { ethErrors, serializeError } = require('eth-json-rpc-errors')
+const { ethErrors } = require('eth-rpc-errors')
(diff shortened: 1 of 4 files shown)


src/utils.js
```diff
@@ -27,7 +26,6 @@ function createErrorMiddleware () {
       if (!error) {
         return done()
       }
-      serializeError(error)
```
rekmarks, line 30: Ref: https://github.com/MetaMask/eth-rpc-errors/blob/master/src/utils.js/#L125

Gudahtt (approved):
LGTM!

PR #84, opened 2020-07-28
6.1.1

rekmarks:
- A couple of package updates


Gudahtt (changes requested):
Changelog entry? Not much to say, but we could at least say that that... dependencies were reduced or updated or something.

rekmarks:
@Gudahtt 🤦 

Changelog updated in: ef09a26

Gudahtt (approved):
LGTM!

PR #85, opened 2020-07-28
Require changelog update on version change

rekmarks:
- Add bash script to CI lint job that exits with `1` if the version changed in `package.json` but the changelog wasn't updated

Sample output of the `diff` command that checks for version changes:
```
> git diff -U0 "${DEFAULT_BRANCH}" -- package.json | grep '"version": "\d\.\d\.\d"'
  -  "version": "6.1.1",
  +  "version": "6.1.2",
```

### DIFF
--- a/.circleci/scripts/require-changelog.sh
+++ b/.circleci/scripts/require-changelog.sh
@@ -1,0 +1,12 @@
+#!/usr/bin/env bash
+
+set -x
+set -e
+set -u
+set -o pipefail
+
+DEFAULT_BRANCH="origin/$(git symbolic-ref refs/remotes/origin/HEAD | sed 's@^refs/remotes/origin/@@')"
+if ( git diff -U0 "${DEFAULT_BRANCH}" -- package.json | grep '"version":' )
+then
+  ( git diff --name-only "${DEFAULT_BRANCH}" | grep CHANGELOG.md ) || exit 1
+fi
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -4,17 +4,14 @@ workflows:
   build-test:
     jobs:
       - prep-deps
-      - test-changelog
       - test-lint:
           requires:
             - prep-deps
       # - test-unit:
       #     requires:
-      #       - test-changelog
       #       - prep-deps
       - all-tests-pass:
           requires:
-            - test-changelog
             - test-lint
             # - test-unit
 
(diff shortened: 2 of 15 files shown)


.circleci/scripts/require-changelog.sh
```diff
@@ -0,0 +1,12 @@
...
+
+set -x
+set -e
+set -u
+set -o pipefail
+
+DEFAULT_BRANCH="origin/$(git symbolic-ref refs/remotes/origin/HEAD | sed 's@^refs/remotes/origin/@@')"
+if ( git diff -U0 "${DEFAULT_BRANCH}" -- package.json | grep '"version": "\d\\.\d\\.\d"' )
+then
+  ( git diff --name-only "${DEFAULT_BRANCH}" | grep CHANGELOG.md ) || exit 1
```
Gudahtt: Is it necessary to determine the "default branch"? I would think something like this would work:

```suggestion
if ( git diff -U0 HEAD^ -- package.json | grep '"version": "\d\\.\d\\.\d"' )
then
  ( git diff --name-only HEAD^ | grep CHANGELOG.md ) || exit 1
```

Gudahtt: Wait, no, this doesn't work 🤦‍♂️ This assumes the PR is just one commit.

rekmarks: ~Good suggestion, that works! 9e8935a~

😅

rekmarks:
Status: TBD if we can find a better way to reference the default branch.

Also, I've tried to make the `grep` matcher more specific. It's currently `grep '"version":'`

Here are things that don't work (all single-quoted):
```
'"version": "\d.\d.\d"' # unescaped periods
'"version": "\d\.\d\.\d"' # single-escaped periods
'"version": "\d\\.\d\\.\d"' # double-escaped periods
'"version": "\d[.]\d[.]\d"' # bracketed periods
```

Is it the `\d` that doesn't work? Is the single-quoted string not actually regex? I've found sources stating quite plainly that it is, and all flavors except the double-escaped periods work locally on my system.

.circleci/scripts/require-changelog.sh
Gudahtt: I wish there was a way to reference the target branch with CircleCI. It looks like this is not a thing, sadly ([though there is an open feature request for it](https://ideas.circleci.com/ideas/CCI-I-894)).

One more reason to switch to GitHub Actions I guess?

rekmarks:
There are better ways to do this with GitHub actions, once we get there.

PR #87, opened 2020-07-29
Remove loglevel

rekmarks:
- Remove `loglevel`
- Replace usage of `loglevel` with `console`

There's not really a point to this dependency; I think it's a leftover from when the various messages were logged in the extension inpage script.

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,7 +31,6 @@
     "is-stream": "^2.0.0",
     "json-rpc-engine": "^5.2.0",
     "json-rpc-middleware-stream": "^2.1.1",
-    "loglevel": "^1.6.1",
     "obj-multiplex": "^1.0.0",
     "obs-store": "^4.0.3",
     "pump": "^3.0.0",
--- a/src/MetamaskInpageProvider.js
+++ b/src/MetamaskInpageProvider.js
@@ -8,7 +8,6 @@ const ObjectMultiplex = require('obj-multiplex')
 const SafeEventEmitter = require('safe-event-emitter')
 const dequal = require('fast-deep-equal')
 const { ethErrors } = require('eth-rpc-errors')
-const log = require('loglevel')
 const { duplex: isDuplex } = require('is-stream')
 
 const messages = require('./messages')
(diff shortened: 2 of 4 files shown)


brad-decker (approved):

whymarrh (dismissed):
I think the benefit of using `loglevel` here (and, I presume, in other dependencies) is that we could/can control what gets outputted in production. Admittedly we're only using `warn` and `error` logging here, but if we did use `info`, `debug`, or `trace` we could easily disable those.

In the extension we set the default level to warn in prod:<sup>[\[1\]](https://github.com/MetaMask/metamask-extension/blob/v8.0.6/app/scripts/inpage.js#L46)</sup>

```js
log.setDefaultLevel(process.env.METAMASK_DEBUG ? 'debug' : 'warn')
```

That, plus the fact that this dependency is (hopefully) de-duplicated in any application it's bundled with means that it's negligible.

All that said, I support removing it, but would suggest we replace its usages with some indirection: we could take a `log` object as a constructor parameter/function argument where needed and allow consumers of the package to pass in something that matches the API/has a similar API. Hard-coding usages of `console` only makes it harder for consumers to influence the logging here.

rekmarks:
@whymarrh it's already impossible for consumers to (in a general way) influence what the provider logs, and I think that's a good thing. If we ever were to use `info` and/or `debug` logging in this package, `loglevel` would of course be useful, but we don't, and I don't know of any plans to do so.

Our only usage of `log` in the extension's inpage script, meanwhile, are a couple of useless `debug` messages that we could also get rid of: https://github.com/MetaMask/metamask-extension/blob/develop/app/scripts/inpage.js/#L76-L80

In short, I believe I understand the benefits you are suggesting, but they seem highly hypothetical to me. If the need to control the logging that happens in production arises, we can always reintroduce `loglevel` at that time.

Gudahtt:
>it's already impossible for consumers to (in a general way) influence what the provider logs

We literally do this today in the inpage script though 😕 That's what `setDefaultLevel` does.

Note that we're not the only consumers of this package. Our recommendation to extensions that want to interact with MetaMask is to use this package, for example.

Gudahtt:
I do agree that we should try to drop this as a dependency (both here and in `inpage.js`), but the idea of passing in the `log` object as a parameter to allow consumers to conditionally disable logs is not without merit.

rekmarks:
Ah, my apologies, confusion over the overloaded (just in my brain?) term "consumers".

I was reading and using it in the sense of API consumers, not consumers of the package. 🤦

‍I'll follow up with a PR that implements my understanding of Whymarrh's suggestion of a constructor argument.

PR #88, opened 2020-07-30
Enable package consumers to specify a logger

rekmarks:
- Add named param `logger` to constructor
  - This object is assigned to a global variable `log`, which is used in place of `console` for all logging calls
  - By default, `logger = console`
- Validate `logger` param
  - Require `typeof 'object' === true` and the following methods:
    - `debug`
    - `error`
    - `info`
    - `log`
    - `trace`
  - Add `logger` validation tests

### DIFF
--- a/src/MetamaskInpageProvider.js
+++ b/src/MetamaskInpageProvider.js
@@ -22,12 +22,22 @@ const {
 
 let log
 
+/**
+ * @typedef {Object} ConsoleLike
+ * @property {function} debug - Like console.debug
+ * @property {function} error - Like console.error
+ * @property {function} info - Like console.info
+ * @property {function} log - Like console.log
+ * @property {function} trace - Like console.trace
+ * @property {function} warn - Like console.warn
+ */
+
 module.exports = class MetamaskInpageProvider extends SafeEventEmitter {
 
   /**
    * @param {Object} connectionStream - A Node.js duplex stream
    * @param {Object} opts - An options bag
-   * @param {Object} [opts.logger] - The logging API to use. Default: console
+   * @param {ConsoleLike} [opts.logger] - The logging API to use. Default: console
    * @param {number} [opts.maxEventListeners] - The maximum number of event
    * listeners. Default: 100
    * @param {boolean} [opts.shouldSendMetadata] - Whether the provider should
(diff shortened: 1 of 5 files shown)


src/MetamaskInpageProvider.js
```diff
@@ -20,19 +20,30 @@ const {
   NOOP,
 } = require('./utils')
 
+let log
```
rekmarks, line 23: Global because we don't want websites to mess with it.

src/MetamaskInpageProvider.js
```diff
@@ -20,19 +20,30 @@ const {
...
     connectionStream,
-    { maxEventListeners = 100, shouldSendMetadata = true } = {},
+    {
+      logger = console,
+      maxEventListeners = 100,
+      shouldSendMetadata = true,
+    } = {},
   ) {
 
+    log = logger
```
brad-decker, line 56: [?]: Shouldn't we do some checking on the provided logger to make sure that it has the expected methods and signatures?

brad-decker (changes requested):

src/MetamaskInpageProvider.js
Gudahtt: Or alternatively, be more specific about the type of the `opts.logger` parameter? (i.e. specify that it must have any functions that we use)

It seems fair to expect consumers to pass in valid parameters if they're typed adequately, but it's not at the moment.

rekmarks: I added validation here: 98298c1

How do we feel about the required methods?

rekmarks: I'll improve the type, too.

rekmarks: `typedef` added in f68e8c3

rekmarks:
`log` and `debug` are aliases in `loglevel`. Should we just require one or the other?
https://npmjs.com/package/loglevel

brad-decker (approved):
Changes LGTM 👍

brad-decker:
We could either manually alias them if one is provided or not the other, or validate that one exists, but if we use one and the other is provided then we have an issue. Technically we only need to validate that the methods used in the library are provided

Gudahtt:
I would consider `console` to be the API to emulate here, not `loglevel`. But I don't think it's important to validate the existence of any methods we don't actually use.

rekmarks:
Fair points, although I decided to keep the properties (which in my mind are the most commonly used) in case we decide to use any of them in the future.

PR #90, opened 2020-08-04
Add prepare script

rekmarks:
Add `prepare` script to help prevent publishing local garbage.

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,7 +7,8 @@
     "test": "jest",
     "coverage": "jest --coverage",
     "lint": "eslint . --ext js,json",
-    "lint:fix": "eslint . --ext js,json --fix"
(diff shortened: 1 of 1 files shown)


brad-decker (approved):

Gudahtt:
>prepare: Run both BEFORE the package is packed and published, on local npm install without any arguments, and when installing git dependencies

I don't think this is what we want here. Did you mean to use `prepublishOnly`?

rekmarks:
Hm, `prepare` gets the job done, but `prepublishOnly` is more specific 🤔

Gudahtt:
The part that seems less than ideal to me is testing/linting upon install. Generally when I run `yarn` I haven't done anything yet, so... I have no reason to lint/test. But yeah, it's not the end of the world.

PR #91, opened 2020-08-04
Replace prepare script with prepublishOnly

rekmarks:
- `prepare` -> `prepublishOnly`
  - Also add check for clean working directory


Gudahtt (approved):
LGTM!

PR #92, opened 2020-08-15
Add types

Jack-Works:
Create `index.d.ts`.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,0 +1,58 @@
+import { EventEmitter } from 'events';
+import { Duplex } from 'stream';
+import { AbstractProvider } from 'web3-core';
+
+export interface MetamaskInpageProviderOptions {
+    /**
+     * The logging API to use.
+     * @default console
+     */
+    logger?: Console;
+    /**
+     * The maximum number of event listeners.
+     * @default 100
+     */
+    maxEventListeners?: number;
+    /**
+     * Whether the provider should send page metadata.
+     * @default true
+     */
+    shouldSendMetadata?: boolean;
+}
+export class MetamaskInpageProvider extends EventEmitter implements AbstractProvider {
+    /**
+     *
+     * @param connectionStream A Node.js duplex stream
+     * @param options An options bag
+     */
+    constructor(connectionStream: Duplex, options?: MetamaskInpageProviderOptions);
+    readonly isMetaMask: true;
+    readonly selectedAddress: string | null;
+    // Help wanted. Is it a string?
+    readonly networkVersion: undefined;
+    // Help wanted. Is it a string?
+    readonly chainId: undefined;
+    get publicConfigStore(): any;
+    isConnected(): boolean;
+    /** Submits an RPC request to MetaMask for the given method, with the given params. Resolves with the result of the method call, or rejects on error. */
+    request(jsonRPCPayload: object): Promise<unknown>;
+    sendAsync(jsonRPCPayload: object, callback: (err: Error | undefined, result: undefined) => void): void;
+}
+/**
+ * Initializes a MetamaskInpageProvider and (optionally) sets it on window.ethereum.
+ * @returns The initialized provider (whether set or not).
+ */
+export function initProvider(
+    opts?: Pick<MetamaskInpageProviderOptions, 'maxEventListeners' | 'shouldSendMetadata'> & {
+        /** A Node.js duplex stream */
+        connectionStream?: Duplex;
+        /** Whether the provider should be set as window.ethereum */
+        shouldSetOnWindow?: boolean;
+    },
+): MetamaskInpageProvider;
+/**
+ * Sets the given provider instance as window.ethereum and dispatches the 'ethereum#initialized' event on window.
+ *
+ * @param providerInstance - The provider instance.
+ */
+export function setGlobalProvider(providerInstance: MetamaskInpageProvider): void;
--- a/index.d.ts
+++ b/index.d.ts
@@ -55,12 +55,17 @@ export function setGlobalProvider(providerInstance: MetamaskInpageProvider): voi
 export interface JsonRpcPayload {
     jsonrpc: string
     method: string
-    params: unknown[]
+    params?: unknown[]
     id?: string | number
 }
 export interface JsonRpcResponse {
     jsonrpc: string
-    id: number
+    id: string | number
     result?: unknown
-    error?: string
+    error?: {
+        code: number
+        message: string
+        data?: unknown
+        stack?: any
+    }
 }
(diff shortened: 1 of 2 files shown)


rekmarks:
Thanks for making this, @Jack-Works! We've started to implement some types for this internally, but never finished the job. We'll be reviewing this in the coming days 👍

index.d.ts
```diff
@@ -0,0 +1,58 @@
+import { EventEmitter } from 'events';
+import { Duplex } from 'stream';
+import { AbstractProvider } from 'web3-core';
```
danfinlay: I see web3-core is LGPL, which could taint the license of our whole project, so we might want to just write this fairly simple part under an even more permissive license.

danfinlay (dismissed):
Overall a great addition, but we should do it without license changes.

index.d.ts
Jack-Works: AbstractProvider is a simple interface I can add it without add web3-core as a dependency. I will do the change later

rekmarks:
@Jack-Works for your reference, @brad-decker (on our team) previously sketched the following types for this package. I'm not suggesting you have to adapt any of this over what you've done, but in case it's useful: https://gist.github.com/rekmarks/1d249cb9d805f8b8ad89467ae961517b

index.d.ts
```diff
@@ -0,0 +1,58 @@
...
+     * @param options An options bag
+     */
+    constructor(connectionStream: Duplex, options?: MetamaskInpageProviderOptions);
+    readonly isMetaMask: true;
+    readonly selectedAddress: string | null;
+    // Help wanted. Is it a string?
+    readonly networkVersion: undefined;
+    // Help wanted. Is it a string?
+    readonly chainId: undefined;
+    get publicConfigStore(): any;
```
rekmarks: We're planning to remove the `publicConfigStore` property, so I'm not sure if we should include it or not 🤔

index.d.ts
```diff
@@ -0,0 +1,58 @@
...
+     * @param connectionStream A Node.js duplex stream
+     * @param options An options bag
+     */
+    constructor(connectionStream: Duplex, options?: MetamaskInpageProviderOptions);
+    readonly isMetaMask: true;
+    readonly selectedAddress: string | null;
+    // Help wanted. Is it a string?
+    readonly networkVersion: undefined;
+    // Help wanted. Is it a string?
+    readonly chainId: undefined;
```
rekmarks: They should be `string | null`

ref: #93

index.d.ts
```diff
@@ -0,0 +1,58 @@
+import { EventEmitter } from 'events';
+import { Duplex } from 'stream';
+import { AbstractProvider } from 'web3-core';
+
+export interface MetamaskInpageProviderOptions {
+    /**
+     * The logging API to use.
+     * @default console
+     */
+    logger?: Console;
```
rekmarks: Could we extend this to allow other loggers? For reference, we should probably turn this into a type: https://github.com/MetaMask/inpage-provider/blob/master/src/MetamaskInpageProvider.js/#L696

index.d.ts
```diff
@@ -0,0 +1,58 @@
...
+    readonly isMetaMask: true;
+    readonly selectedAddress: string | null;
+    // Help wanted. Is it a string?
+    readonly networkVersion: undefined;
+    // Help wanted. Is it a string?
+    readonly chainId: undefined;
+    get publicConfigStore(): any;
+    isConnected(): boolean;
+    /** Submits an RPC request to MetaMask for the given method, with the given params. Resolves with the result of the method call, or rejects on error. */
+    request(jsonRPCPayload: object): Promise<unknown>;
```
rekmarks: The `request` parameter should follow the `RequestArguments` type defined in EIP-1193: https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1193.md#request

I also think it's in `web3-core`.

index.d.ts
```diff
@@ -0,0 +1,58 @@
...
+    sendAsync(jsonRPCPayload: object, callback: (err: Error | undefined, result: undefined) => void): void;
```
rekmarks: The `jsonRPCPayload` should be typed like it is in `web3-core`.

index.d.ts
```diff
@@ -0,0 +1,66 @@
...
+export function setGlobalProvider(providerInstance: MetamaskInpageProvider): void
+export interface JsonRpcPayload {
+    jsonrpc: string
+    method: string
+    params: unknown[]
+    id?: string | number
+}
+export interface JsonRpcResponse {
+    jsonrpc: string
+    id: number
```
rekmarks:
```suggestion
    id: string | number
```

index.d.ts
```diff
@@ -0,0 +1,66 @@
...
+/**
+ * Sets the given provider instance as window.ethereum and dispatches the 'ethereum#initialized' event on window.
+ *
+ * @param providerInstance - The provider instance.
+ */
+export function setGlobalProvider(providerInstance: MetamaskInpageProvider): void
+export interface JsonRpcPayload {
+    jsonrpc: string
+    method: string
+    params: unknown[]
```
rekmarks: `params` are not always required.
```suggestion
    params?: unknown[]
```

index.d.ts
```diff
@@ -0,0 +1,66 @@
...
+    id?: string | number
+}
+export interface JsonRpcResponse {
+    jsonrpc: string
+    id: number
+    result?: unknown
+    error?: string
```
rekmarks: All RPC errors (from `sendAsync`, `request`, `send`) follow this interface, without the `stack`: https://github.com/MetaMask/eth-rpc-errors/blob/master/%40types/index.d.ts#L2-L7

brad-decker (dismissed):
LGTM, good work!

index.d.ts
```diff
@@ -0,0 +1,71 @@
...
+}
+export interface JsonRpcResponse {
+    jsonrpc: string
+    id: string | number
+    result?: unknown
+    error?: {
+        code: number
+        message: string
+        data?: unknown
+        stack?: any
```
rekmarks: The `stack` will never be present here in practice:
```suggestion
```

index.d.ts
```diff
@@ -0,0 +1,71 @@
...
+     */
+    constructor(connectionStream: Duplex, options?: MetamaskInpageProviderOptions)
+    readonly isMetaMask: true
+    readonly selectedAddress: string | null
+    readonly networkVersion: string | null
+    readonly chainId: string | undefined
+    isConnected(): boolean
+    sendAsync(payload: JsonRpcPayload, callback: (error: Error | null, result?: JsonRpcResponse) => void): void
+    /** Submits an RPC request to MetaMask for the given method, with the given params. Resolves with the result of the method call, or rejects on error. */
+    request(args: JsonRpcPayload): Promise<any>
```
rekmarks: We should prefer `unknown`:
```suggestion
    request(args: JsonRpcPayload): Promise<unknown>
```

rekmarks (changes requested):
Just a couple more pieces of minor changes, then we should be good to go!

rekmarks (approved):
LGTM!

Great work @Jack-Works. Many thanks!

PR #93, opened 2020-08-16
Set networkVersion and chainId defaults to null

rekmarks:
Sets the `networkVersion` and `chainId` properties to `null` by default and if falsy values are returned from the background.

Previously, they were `undefined` by default, while `selectedAddress` was `null`. This change is more in keeping with the use of `null` for values that are intentionally not set/"undefined", which is the case for `networkVersion` and `chainId` on boot and if we get falsy values from the background.


whymarrh (approved):
LGTM

PR #97, opened 2020-08-19
Touch up types

rekmarks:
- Add full docstrings to declarations file
  - The docstrings in the JavaScript source are not picked up by TypeScript, at least in VS Code.
  - Make minor edits to source docstrings in process.
- Add typings for all `send` method signatures
- Use `RequestArguments` interface for `request` method
- Require `opts.connectionStream` param for `initProvider`
- Update `jsonrpc` property on `JsonRpcRequest` and `JsonRpcResponse` interfaces
- Update `JsonRpcResponse` interface to type with either `result` or `error` property
- Lint types (per our config, but without adding dependencies)
  - Basically adds spacing and semicolons

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -25,53 +25,88 @@ export interface MetamaskInpageProviderOptions {
 export class MetamaskInpageProvider extends EventEmitter {
 
   /**
-   * @param connectionStream A Node.js duplex stream
-   * @param options An options bag
+   * @param connectionStream - A Node.js duplex stream.
+   * @param options - An options bag.
    */
   constructor (connectionStream: Duplex, options?: MetamaskInpageProviderOptions);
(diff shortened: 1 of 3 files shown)


brad-decker (approved):

index.d.ts
```diff
@@ -1,70 +1,174 @@
...
+  /**
+   * The user's currently selected Ethereum address.
+   * If null, MetaMask is either locked or the user has not permitted any
+   * addresses to be viewed.
+   */
+  readonly selectedAddress: string | null;
+
+  /**
+   * The network ID of the currently connected Ethereum chain.
+   * @deprecated Use {@link chainId} instead.
```
brad-decker, line 88: this is daf.

PR #101, opened 2020-09-08
Fix MetaMask casing

rekmarks:
`s/Metamask/MetaMask/g`

Fixes MetaMask camel casing for everything: files, exports, variables etc.

A bit frivolous, perhaps, but the Brand demands it.


whymarrh (approved):
A breaking change but a good one!

PR #103, opened 2020-09-08
Add Node types

rekmarks:
- `@types/node@^14.6.4`
  - Either the consumer has to bring this themselves or we do


whymarrh:
There are cases where the consumer isn't written in TypeScript where this dependency isn't welcome. Can we continue to leave this out?

rekmarks:
We're not gonna do it!

PR #108, opened 2020-11-24
Document type dependencies

rekmarks:
Documents type dependencies since we're not shipping `@types/node`.


Gudahtt (approved):
Seems like an improvement!

I wish there was a way to add _optional_ peer dependencies. Yarn v2 specifies a way, but it would mess up downstream `npm` projects if we relied on that. Oh well!

PR #109, opened 2020-11-24
Replace public config store with RPC methods/notifications

rekmarks:
- Deletes the public config stream, `_publicConfigStore`, and `publicConfigStore` getter
- Gets state previously received through public config stream with RPC method and notifications
  - New RPC Methods
    - `wallet_getProviderState`
  - New RPC Notifications
    - `wallet_unlockStateChanged`
    - `wallet_chainChanged`

Corresponding PR in extension: https://github.com/MetaMask/metamask-extension/pull/8640

CI: https://app.circleci.com/pipelines/github/MetaMask/inpage-provider?branch=delete-public-config

### DIFF
--- a/src/MetaMaskInpageProvider.js
+++ b/src/MetaMaskInpageProvider.js
@@ -152,12 +152,12 @@ module.exports = class MetaMaskInpageProvider extends SafeEventEmitter {
     jsonRpcConnection.events.on('notification', (payload) => {
       const { method, params, result } = payload
 
-      if (method === 'wallet_accountsChanged') {
+      if (method === 'metamask_accountsChanged') {
         this._handleAccountsChanged(result)
 
-      } else if (method === 'wallet_unlockStateChanged') {
+      } else if (method === 'metamask_unlockStateChanged') {
         this._handleUnlockStateChanged(result)
-      } else if (method === 'wallet_chainChanged') {
+      } else if (method === 'metamask_chainChanged') {
         this._handleChainChanged(result)
       } else if (EMITTED_NOTIFICATIONS.includes(method)) {
         this.emit('notification', payload) // deprecated
@@ -304,14 +304,14 @@ module.exports = class MetaMaskInpageProvider extends SafeEventEmitter {
 
   /**
    * Constructor helper.
-   * Populates initial state by calling 'wallet_getProviderState' and
+   * Populates initial state by calling 'metamask_getProviderState' and
    * 'eth_accounts', and emits necessary events.
    *
    * @private
    */
   _initializeState () {
     Promise.all([
-      this.request({ method: 'wallet_getProviderState' }),
+      this.request({ method: 'metamask_getProviderState' }),
       this.request({ method: 'eth_accounts' }),
     ])
       .then(([state, accounts]) => {
--- a/src/MetaMaskInpageProvider.js
+++ b/src/MetaMaskInpageProvider.js
@@ -1,8 +1,6 @@
 const pump = require('pump')
 const { JsonRpcEngine, createIdRemapMiddleware } = require('json-rpc-engine')
 const createJsonRpcStream = require('json-rpc-middleware-stream')
-const ObservableStore = require('obs-store')
-const asStream = require('obs-store/lib/asStream')
 const ObjectMultiplex = require('obj-multiplex')
 const SafeEventEmitter = require('safe-event-emitter')
 const dequal = require('fast-deep-equal')
@@ -50,7 +48,6 @@ module.exports = class MetaMaskInpageProvider extends SafeEventEmitter {
       shouldSendMetadata = true,
     } = {},
   ) {
-
     validateLoggerObject(logger)
     log = logger
 
@@ -87,13 +84,10 @@ module.exports = class MetaMaskInpageProvider extends SafeEventEmitter {
           networkChanged: false,
           notification: false,
         },
-        // misc
-        // TODO:deprecation:remove
-        publicConfigStore: false,
       },
-      isConnected: undefined,
-      accounts: undefined,
-      isUnlocked: undefined,
+      accounts: null,
+      isConnected: false,
+      isUnlocked: false,
     }
 
     this._metamask = this._getExperimentalApi()
@@ -105,6 +99,8 @@ module.exports = class MetaMaskInpageProvider extends SafeEventEmitter {
 
     // bind functions (to prevent consumers from making unbound calls)
     this._handleAccountsChanged = this._handleAccountsChanged.bind(this)
+    this._handleChainChanged = this._handleChainChanged.bind(this)
+    this._handleUnlockStateChanged = this._handleUnlockStateChanged.bind(this)
     this._handleDisconnect = this._handleDisconnect.bind(this)
     this._sendSync = this._sendSync.bind(this)
     this._rpcRequest = this._rpcRequest.bind(this)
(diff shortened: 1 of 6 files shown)


Gudahtt:
How does this relate to the breaking changes? Did you just want to bundle it with them for the sake of avoiding another breaking change release, or is it functionally tied to them in some way?

rekmarks:
Ah, we're bundling this breaking change with all the rest just for the sake of convenience!

Gudahtt (commented):
Should we be using `metamask_` here instead of `wallet_`? :thinking: 

I had thought we were, as a general rule, using `metamask_` for methods that are internal-only or that we have no plans to standardize, and `wallet_` for things that _might_ be standardized at some point. These methods seem to me of internal relevance only, at least at the moment.

src/MetaMaskInpageProvider.js
```diff
@@ -356,16 +302,50 @@ module.exports = class MetaMaskInpageProvider extends SafeEventEmitter {
...
+   * 'eth_accounts', and emits necessary events.
+   *
+   * @private
+   */
+  _initializeState () {
+    Promise.all([
+      this.request({ method: 'wallet_getProviderState' }),
+      this.request({ method: 'eth_accounts' }),
+    ])
+      .then(([state, accounts]) => {
```
Gudahtt: Nit: `async / await` should be preferred over `.then / .catch`

rekmarks:
@Gudahtt, good point. I had considered that but never executed. Done in: 2fe974d

src/siteMetadata.js
```diff
@@ -18,7 +18,7 @@ async function sendSiteMetadata (engine, log) {
     // call engine.handle directly to avoid normal RPC request handling
     engine.handle(
       {
-        method: 'wallet_sendDomainMetadata',
+        method: 'metamask_sendDomainMetadata',
```
Gudahtt, line 21: This seems out of scope for this PR, and `wallet_` might be the more appropriate prefix for this method anyway. We had talked about encouraging dapps to set their own metadata. So we might want to consider documenting this as-is.

src/MetaMaskInpageProvider.js
```diff
@@ -485,19 +537,13 @@ module.exports = class MetaMaskInpageProvider extends SafeEventEmitter {
          * @returns {Promise<boolean>} - Promise resolving to true if MetaMask is currently unlocked
          */
         isUnlocked: async () => {
-          if (this._state.isUnlocked === undefined) {
```
Gudahtt, line 488: This is an unfortunate change in behaviour. This used to block on determining the actual unlock status of the wallet, but now it returns `false` right away if called during initialization, even if the wallet is unlocked.

src/MetaMaskInpageProvider.js
```diff
@@ -459,8 +441,77 @@ module.exports = class MetaMaskInpageProvider extends SafeEventEmitter {
...
+
+    if (isUnlocked !== this._state.isUnlocked) {
+      this._state.isUnlocked = isUnlocked
+
+      if (isUnlocked) {
+
+        // this will get the exposed accounts, if any
+        try {
+          this._rpcRequest(
+            { method: 'eth_accounts', params: [] },
```
Gudahtt, line 503: Could we bundle the accounts with the unlock notification, to avoid this second call? It would simplify things a great deal. I think it would let us remove the entire `isInternal` parameter altogether, and that `'eth_accounts' unexpectedly updated accounts` warning

Gudahtt: I guess this can wait for a subsequent PR either way though.

src/MetaMaskInpageProvider.js
```diff
@@ -459,8 +441,77 @@ module.exports = class MetaMaskInpageProvider extends SafeEventEmitter {
...
+      )
+      return
+    }
+
+    if (chainId !== this.chainId || networkVersion !== this.networkVersion) {
+      this.chainId = chainId
+      this.emit('chainChanged', this.chainId)
+
+      this.networkVersion = networkVersion
+      this.emit('networkChanged', this.networkVersion)
```
Gudahtt, line 477: This would re-emit `networkChanged` even if it remained the same (e.g. if only the `chainId` changed). Was this intentional? It is a change in behaviour from before.

src/MetaMaskInpageProvider.js
```diff
@@ -356,16 +302,50 @@ module.exports = class MetaMaskInpageProvider extends SafeEventEmitter {
...
+   * Constructor helper.
+   * Populates initial state by calling 'metamask_getProviderState' and
+   * 'eth_accounts', and emits necessary events.
+   *
+   * @private
+   */
+  _initializeState () {
+    Promise.all([
+      this.request({ method: 'metamask_getProviderState' }),
+      this.request({ method: 'eth_accounts' }),
```
Gudahtt: Any reason `eth_accounts` is a separate call, rather than being included with the provider state?

Gudahtt: Not that I object to us doing this for now - we can optimize this later if we feel like it. Just wondering if there was a reason beyond that this is closer to the old behaviour.

src/MetaMaskInpageProvider.js
rekmarks: Hm, the only case where the chain ID but _not_ the network ID changes is for the few cases where chains share the latter but not the former. "changes" here we'll take to mean "becomes a different value in state."

In those cases, I think it's fair to say that "the network changed" even though the network ID didn't. I'm really agnostic here, since we consider the the network ID and its related event deprecated.

src/MetaMaskInpageProvider.js
rekmarks: Hm, this is actually very feasible now. Let's do that in a follow-up!

src/siteMetadata.js
rekmarks: I always considered it to be kind of a hack, personally, and I think there's more work to be done on specifying the method before we commit to publishing its interface.

As to the out of scope, yeah kind of, but while we're here? 😅

src/MetaMaskInpageProvider.js
rekmarks: Done here: 79128b4

src/MetaMaskInpageProvider.js
rekmarks: Done here: 79128b4

The `getProviderState` implementation was updated on the extension side as well. I'll thoroughly test that with this version of the inpage provider before opening the extension PR.

src/MetaMaskInpageProvider.js
rekmarks: I restored that behavior here: c0b32ae

src/MetaMaskInpageProvider.js
rekmarks: EIP-1193 doesn't specify what to do, so it's really a judgment call: https://eips.ethereum.org/EIPS/eip-1193#networkchanged-deprecated

I'd say that preserving existing behavior is best, but the network _has actually changed_ whenever the chain ID value but not the network ID value changes.

Gudahtt: Yeah. If not for legacy reasons, I'd agree that re-emitting makes the most sense. But since this behaviour might be relied upon, and since we're deprecating it anyway, I think it'd be better to preserve the old behaviour.

src/siteMetadata.js
Gudahtt: >I think there's more work to be done on specifying the method before we commit to publishing its interface.

Makes sense. I hadn't really thought through whether it was ready to be public-facing in its current form.

It would have been better if this was `metamask_` in the first place., but... it still seems better to leave it as is for now :man_shrugging: Technically dapps might already be sending this? Also the less provider changes we make, the less broken things with embedded providers will be (e.g. extensions, and dapps that import our provider as a workaround to that Firefox CSP bug). But I guess I won't block on it. Embedded providers will be _fairly broken_ either way.

src/MetaMaskInpageProvider.js
rekmarks: Aight, I restored the behavior here: 94d84d6

src/siteMetadata.js
Gudahtt: Actually... maybe we can preserve the public config store in the extension to support old embedded providers, at least for a short while? :thinking: I don't know how feasible this is - I'll have to look more closely at the extension PRs.

Gudahtt (approved):
LGTM!

rekmarks:
CI passes: https://app.circleci.com/pipelines/github/MetaMask/inpage-provider/388/workflows/e6397b45-6ba6-4d5b-ae4d-f518d981cb28

PR #113, opened 2020-11-28
Add web3 shim

rekmarks:
Adds a function for shimming `window.web3.currentProvider`, which `initProvider` can be configured to call.

### DIFF
--- a/src/shimWeb3.js
+++ b/src/shimWeb3.js
@@ -25,11 +25,17 @@ module.exports = function shimWeb3 (provider) {
           }
           return Reflect.get(target, property, ...args)
         },
+        set: (...args) => {
+          console.warn(
+            'You are accessing the MetaMask window.web3 shim. This object is deprecated; use window.ethereum instead. For details, see: https://docs.metamask.io/guide/provider-migration.html#replacing-window-web3',
+          )
+          return Reflect.set(...args)
+        },
       },
     )
 
     Object.defineProperty(window, 'web3', {
-      value: Object.freeze(web3Shim),
+      value: web3Shim,
       enumerable: false,
       configurable: true,
       writable: true,
--- a/src/shimWeb3.js
+++ b/src/shimWeb3.js
@@ -1,0 +1,38 @@
+/**
+ * If no existing window.web3 is found, this function injects a web3 "shim" to
+ * not break dapps that rely on window.web3.currentProvider.
+ *
+ * @param {import('./MetaMaskInpageProvider')} provider - The provider to set as window.web3.currentProvider.
+ */
+module.exports = function shimWeb3 (provider) {
+  if (!window.web3) {
+    const SHIM_IDENTIFIER = '__isMetaMaskShim__'
+    const web3Shim = new Proxy(
+      {
+        currentProvider: provider,
+        [SHIM_IDENTIFIER]: true,
+      },
+      {
+        get: (target, property, ...args) => {
+          if (property === 'currentProvider') {
+            console.warn(
+              'You are accessing the MetaMask window.web3.currentProvider shim. Just use window.ethereum instead. For details, see: https://docs.metamask.io/guide/provider-migration.html#replacing-window-web3',
+            )
+          } else if (property !== SHIM_IDENTIFIER) {
+            console.error(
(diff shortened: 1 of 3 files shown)


src/shimWeb3.js
```diff
@@ -0,0 +1,38 @@
...
+    const web3Shim = new Proxy(
+      {
+        currentProvider: provider,
+        [SHIM_IDENTIFIER]: true,
+      },
+      {
+        get: (target, property, ...args) => {
+          if (property === 'currentProvider') {
+            console.warn(
+              'You are accessing the MetaMask window.web3 shim. Just use window.ethereum instead. For details, see: https://docs.metamask.io/guide/ethereum-provider.html',
```
Gudahtt: This might be confusing for some devs :thinking: It doesn't say _why_ they should use `window.ethereum` instead. Maybe we can call this "deprecated"? That's a word people understand.

src/shimWeb3.js
```diff
@@ -0,0 +1,38 @@
...
+            )
+          } else if (property !== SHIM_IDENTIFIER) {
+            console.error(
+              `You are accessing the MetaMask window.web3 shim. The property '${property}' does not exist. For details, see: https://docs.metamask.io/guide/ethereum-provider.html`,
```
Gudahtt: This could be pretty confusing. If I expected this to be a "real" `web3` instance, this would leave me bewildered (What is "the MetaMask window.web3 shim"? Why doesn't the `eth` property exist? How do I use this?). The docs don't explain what the shim is or why it exists.

Maybe we need another docs section :thinking: Or maybe it would suffice to say something like "MetaMask no longer injects `window.web3`"

rekmarks: I updated both error messages and directed the user to the provider migration guide. I think it already goes a long way, but we can add a subsection describing what `window.web3.currentProvider` is. Would that suffice?

src/shimWeb3.js
```diff
@@ -0,0 +1,38 @@
...
+    const web3Shim = new Proxy(
+      {
+        currentProvider: provider,
+        [SHIM_IDENTIFIER]: true,
+      },
+      {
+        get: (target, property, ...args) => {
+          if (property === 'currentProvider') {
+            console.warn(
+              'You are accessing the MetaMask window.web3.currentProvider shim. Just use window.ethereum instead. For details, see: https://docs.metamask.io/guide/provider-migration.html#replacing-window-web3',
```
Gudahtt: How does this sound?
```suggestion
              'You are accessing the MetaMask window.web3.currentProvider shim. This property is deprecated; use window.ethereum instead. For details, see: https://docs.metamask.io/guide/provider-migration.html#replacing-window-web3',
```

I have an aversion to using the word "just" in error messages because it can feel dismissive of people's confusion, and I think calling this "deprecated" better communicates the status of this property.

src/shimWeb3.js
Gudahtt: This seems OK to me!

Gudahtt: Oh and yes, a subsection describing the shim would be very helpful as well.

src/shimWeb3.js
```diff
@@ -0,0 +1,38 @@
...
+              `MetaMask no longer injects web3. For details, see: https://docs.metamask.io/guide/provider-migration.html#replacing-window-web3`,
+            )
+          }
+          return Reflect.get(target, property, ...args)
+        },
+      },
+    )
+
+    Object.defineProperty(window, 'web3', {
+      value: Object.freeze(web3Shim),
```
Gudahtt: This might make some dapps blow up :thinking:  In strict mode, mutating a frozen object will throw a `TypeError`.

I don't really know why a dapp would mutate web3, or what we should do in this scenario. Throwing an error might help dapp devs track down this problem. Have you encountered any dapps that mutate web3?

Gudahtt (dismissed):
LGTM!

brad-decker (approved):
LGTM as well.

rekmarks:
CI passes: https://app.circleci.com/pipelines/github/MetaMask/inpage-provider/384/workflows/421bb2f8-9a51-41ba-9546-d7727c1b4d55

PR #116, opened 2020-12-03
Add logger argument to initializeProvider, log web3 shim usage

rekmarks:
- Add `logger` argument to `initializeProvider`
  - This is now passed on to the pre-existing `logger` argument of the inpage provider constructor
- Calls the `metamask_logWeb3ShimUsage` RPC method whenever a property other than the shim id/flag or `currentProvider` is accessed.
  - The RPC method is implemented here: https://github.com/MetaMask/metamask-extension/pull/9156

### DIFF
--- a/src/shimWeb3.js
+++ b/src/shimWeb3.js
@@ -3,8 +3,9 @@
  * not break dapps that rely on window.web3.currentProvider.
  *
  * @param {import('./MetaMaskInpageProvider')} provider - The provider to set as window.web3.currentProvider.
+ * @param {typeof console} log - The logging API to use.
  */
-module.exports = function shimWeb3 (provider) {
+module.exports = function shimWeb3 (provider, log = console) {
   if (!window.web3) {
     const SHIM_IDENTIFIER = '__isMetaMaskShim__'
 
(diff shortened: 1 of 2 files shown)


src/shimWeb3.js
```diff
@@ -28,6 +28,8 @@ module.exports = function shimWeb3 (provider) {
             console.error(
               `MetaMask no longer injects web3. For details, see: https://docs.metamask.io/guide/provider-migration.html#replacing-window-web3`,
             )
+            provider.request({ method: 'metamask_logWeb3ShimUsage' })
+              .catch(() => undefined)
```
rekmarks: I debated whether to add any logging. Regarding the use of `.catch`, we don't care about the stack trace, and I don't think it's worth the effort to enable `try`/`catch` here.

Gudahtt: If we don't log something, how would we know whether it was broken?

rekmarks: Well, it'll only show up on the user's end, and it could fail for reasons we don't care about. For the user to notify us, we need some kind of call to action. Should we just ask them report it?

rekmarks: Any unexpected errors would show up in the MetaMask background, so we could add logging there.

Gudahtt: I just mean that without _some_ signal, even the simplest of errors can result in painfully difficult debugging. Adding a `console.debug` log would at least let a developer know that something atypical is happening. That could be a useful clue, even without a Sentry error report or call-to-action or anything like that.

rekmarks: Yeah, that's fair. I updated the logging and added a `log.debug` here: 67a8cf6

Gudahtt (approved):
LGTM!

PR #117, opened 2020-12-04
Only emit state change events after initialization

### LINKED ISSUE #110: chainChanged emits on page load
**Describe the bug**
Sometimes `chainChanged` message emits on every page load. So if I have this line in my code: `ethereum.on('chainChanged', (_chainId) => window.location.reload());` then user gets endless page refresh loop.

**To Reproduce**
I couldn't reproduce this, but one of our user shared his screen and I saw this endless page refresh loop. I added `  console.log('chainChanged');` to that event callback, and console showed this log every time before page refreshed.

**Expected behavior**
`chainChanged` event should be sent only on chain change, not on page load

**Browser details**
 - OS: OS X
 - Browser: last Chrome
 - MetaMask Version: 8.0.8

rekmarks:
It was reported in #110 that `chainChanged` (and by extension, `networkChanged`) is sometimes emitted for the first time after page load, causing a reload loop for those handling chain changes by reloading the page. When the `chainId` is set on initialization, the chain has not changed, and `chainChanged` should not be emitted. The same can be said for `accountsChanged`, which currently would emit under the same circumstances.

This PR ensures that `chainChanged`, `networkChanged`, and `accountsChanged` are only emitted after the provider has been initialized.

Fixes #110

### DIFF
--- a/src/MetaMaskInpageProvider.js
+++ b/src/MetaMaskInpageProvider.js
@@ -532,6 +532,7 @@ module.exports = class MetaMaskInpageProvider extends SafeEventEmitter {
         this.selectedAddress = _accounts[0] || null
       }
 
+      // finally, after all state has been updated, emit the event
       if (this._state.initialized) {
         this.emit('accountsChanged', _accounts)
       }
--- a/src/MetaMaskInpageProvider.js
+++ b/src/MetaMaskInpageProvider.js
@@ -438,8 +438,10 @@ module.exports = class MetaMaskInpageProvider extends SafeEventEmitter {
         this.selectedAddress = _accounts[0] || null
       }
 
-      // only emit the event once all state has been updated
-      this.emit('accountsChanged', _accounts)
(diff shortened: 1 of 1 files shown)


brad-decker:
LGTM, after merge conflicts @rekmarks

src/MetaMaskInpageProvider.js
```diff
@@ -528,8 +532,9 @@ module.exports = class MetaMaskInpageProvider extends SafeEventEmitter {
         this.selectedAddress = _accounts[0] || null
       }
 
-      // only emit the event once all state has been updated
-      this.emit('accountsChanged', _accounts)
+      if (this._state.initialized) {
```
rekmarks, line 540:
```suggestion
      // finally, after all state has been updated, emit the event
      if (this._state.initialized) {
```

brad-decker (approved):

PR #118, opened 2020-12-04
Convert logger to instance variable

rekmarks:
The `log` variable in `MetaMaskProvider.js` was set in that file's global scope. If multiple providers were initialized in the same JavaScript process, giving different loggers to each would change the logger across all of them.

This PR converts the `log` variable to a `_log` property on the provider.


Gudahtt (approved):
LGTM, good catch!

PR #119, opened 2020-12-04
Ensure metadata is sent even if document is already fully loaded

rekmarks:
The metadata sending functionality assumed synchronous injection, which we shouldn't. This PR ensures metadata is sent event if the page is already fully loaded when the constructor runs.


brad-decker (approved):
LGTM

PR #133, opened 2021-01-12
Reduce web3 shim console noise

rekmarks:
- Only log the distinct web3 shim warnings once
- Only send web3 shim usage notification to background once


Gudahtt (approved):
LGTM!

PR #142, opened 2021-04-11
Repo standardization

ryanml:
* Enforcing minimum node version 12 via .nvmrc, and the package file
* Adding basic .gitattributes file
* Adding basic .editorconfig file

Related: https://github.com/MetaMask/MetaMask-planning/issues/32


test/MetaMaskInpageProvider.misc.test.js
```diff
@@ -61,7 +61,7 @@ describe('MetaMaskInpageProvider: Miscellanea', () => {
 
       expect(
         () => new MetaMaskInpageProvider(stream, null),
-      ).toThrow('Cannot destructure property `jsonRpcStreamName` of \'undefined\' or \'null\'');
+      ).toThrow('Cannot read property \'jsonRpcStreamName\' of null');
```
ryanml, line 64: This _might_ be a harmless side effect of the node 10 -> 12 update, but I want to double confirm, so this is a tentative change for now

Gudahtt: It looks like a difference between v10 and v12, specifically in the case when you're using a destructured function parameter with a default value, and you invoke it with `null` for that parameter.

```
const test = ({ bar = 'foo' } = {}) => console.log(bar)
test(null);
```
^ That example throws this on v10: `TypeError: Cannot destructure property `bar` of 'undefined' or 'null'.`
but throws this on v12: `Uncaught TypeError: Cannot read property 'bar' of null`

Gudahtt (approved):
LGTM!

PR #144, opened 2021-04-22
fix: refactor to BaseProvider

shanejonas:


src/BaseProvider.ts
```diff
@@ -0,0 +1,667 @@
...
+    enable: boolean;
+    experimentalMethods: boolean;
+    send: boolean;
+    // events
+    events: {
+      close: boolean;
+      data: boolean;
+      networkChanged: boolean;
+      notification: boolean;
+    };
```
rekmarks: All of these warnings and their associated events and properties should be moved to the `MetaMaskInpageProvider`.

I suppose we have add types such that subclasses can extended the state of `BaseProvider`.

src/BaseProvider.ts
```diff
@@ -0,0 +1,667 @@
...
+
+  /**
+   * Indicating that this provider is a MetaMask provider.
+   */
+  public readonly isMetaMask: true;
+
+  /**
+   * Experimental methods can be found here.
+   */
+  public readonly _metamask: ReturnType<BaseProvider['_getExperimentalApi']>;
```
rekmarks: As these are not part of EIP-1193, I think all of them should be moved to `InpageProvider`. Even `isMetaMask` will not be relevant in all contexts.

The reason we've kept `_metamask` and its methods around is because of `_metamask.isUnlocked`, but it just struck, that's a trivial RPC method that we should implement.

src/BaseProvider.ts
```diff
@@ -0,0 +1,667 @@
...
+   * @param options.maxEventListeners - The maximum number of event
+   * listeners. Default: 100
+   */
+  constructor(
+    connectionStream: typeof Duplex,
+    {
+      jsonRpcStreamName = 'metamask-provider',
+      logger = console,
+      maxEventListeners = 100,
+      shouldSendMetadata = true,
```
rekmarks: `shouldSendMetadata` and its functionality are specific to `InpageProvider` and should be moved there.

src/BaseProvider.ts
```diff
@@ -0,0 +1,667 @@
...
+    if (
+      typeof maxEventListeners !== 'number' ||
+      typeof shouldSendMetadata !== 'boolean'
+    ) {
+      throw new Error(messages.errors.invalidOptions(
+        maxEventListeners, shouldSendMetadata,
+      ));
+    }
+
+    validateLoggerObject(logger);
```
rekmarks: This input validation is a pre-TypeScript hangover that we can probably take the opportunity to get rid of.

The only thing to keep would be `validateLoggerObject` if `ConsoleLike` isn't a good enough type.

src/BaseProvider.ts
```diff
@@ -0,0 +1,667 @@
...
+    {
+      jsonRpcStreamName = 'metamask-provider',
+      logger = console,
+      maxEventListeners = 100,
+      shouldSendMetadata = true,
+    }: BaseProviderOptions = {},
+  ) {
+    if (!isDuplex(connectionStream)) {
+      throw new Error(messages.errors.invalidDuplexStream());
+    }
```
rekmarks, line 130: I'd suggest we get rid of this input validation as well, but Node stream types can be extremely painful to work with, and this is a basically foolproof way to cause invalid streams at runtime.

src/BaseProvider.ts
```diff
@@ -0,0 +1,667 @@
...
+    const mux = new ObjectMultiplex();
+    pump(
+      connectionStream,
+      mux as unknown as Duplex,
+      connectionStream,
+      this._handleStreamDisconnect.bind(this, 'MetaMask'),
+    );
+
+    // ignore phishing warning message (handled elsewhere)
+    mux.ignoreStream('phishing');
```
rekmarks: Should be moved to `InpageProvider`.

src/BaseProvider.ts
```diff
@@ -0,0 +1,667 @@
...
+      if (document.readyState === 'complete') {
+        sendSiteMetadata(this._rpcEngine, this._log);
+      } else {
+        const domContentLoadedHandler = () => {
+          sendSiteMetadata(this._rpcEngine, this._log);
+          window.removeEventListener('DOMContentLoaded', domContentLoadedHandler);
+        };
+        window.addEventListener('DOMContentLoaded', domContentLoadedHandler);
+      }
+    }
```
rekmarks: This should be moved to `InpageProvider`.

src/BaseProvider.ts
```diff
@@ -0,0 +1,667 @@
...
+  }
+
+  /**
+   * Constructor helper.
+   * Gets experimental _metamask API as Proxy, so that we can warn consumers
+   * about its experiment nature.
+   */
+  protected _getExperimentalApi() {
+    return new Proxy(
```
rekmarks: As part of the removal of the `_metamask` property, this should also be moved to `InpageProvider`.

src/initializeProvider.ts
```diff
@@ -1,8 +1,9 @@
 import { Duplex } from 'stream';
```
rekmarks: This file should probably be renamed `initializeInpageProvider`, because it assumes a browser environment.

rekmarks (changes requested):
Great addition with the tests!

I haven't reviewed them yet, because there's a fair amount in `BaseProvider` that should be moved to `MetaMaskInpageProvider`. `BaseProvider` should only include the non-deprecated events, properties, and methods defined in [EIP-1193](https://eips.ethereum.org/EIPS/eip-1193).

src/MetaMaskInpageProvider.ts
```diff
@@ -149,144 +88,42 @@ export default class MetaMaskInpageProvider extends SafeEventEmitter {
...
     // ignore phishing warning message (handled elsewhere)
-    mux.ignoreStream('phishing');
-
-    // setup own event listeners
-
-    // EIP-1193 connect
-    this.on('connect', () => {
-      this._state.isConnected = true;
-    });
+    // mux.ignoreStream('phishing');
```
shanejonas: dont have access to `mux` here, not sure where to put this

src/MetaMaskInpageProvider.ts
```diff
@@ -506,245 +201,29 @@ export default class MetaMaskInpageProvider extends SafeEventEmitter {
...
-    }
-  }
-
   /**
    * Warns of deprecation for the given event, if applicable.
    */
-  private _warnOfDeprecation(eventName: string): void {
-    if (this._state.sentWarnings.events[eventName as WarningEventName] === false) {
+  protected _warnOfDeprecation(eventName: string): void {
+    if (this.sentWarnings && this.sentWarnings.events[eventName as WarningEventName] === false) {
```
shanejonas: having this guard `this.sentWarnings &&` prevents this from running in the constructor, which is not ideal. typescript initializes its default properties after the base constructor runs

src/MetaMaskInpageProvider.ts
Gudahtt: Hmm. So this isn't needed for the inpage provider at all, because the phishing stream is intercepted by the concentscript. It doesn't get here at all.

But this interception doesn't happen with the extension provider :thinking: But maybe we can just... not send phishing events to other extensions specifically? We might be able to get by without this with an extension background change.

rekmarks: Yeah, this line has been unnecessary for a long time. We can just get rid of it!

src/BaseProvider.ts
```diff
@@ -0,0 +1,531 @@
...
+   * @param options.maxEventListeners - The maximum number of event
+   * listeners. Default: 100
+   */
+  constructor(
+    connectionStream: typeof Duplex,
+    {
+      jsonRpcStreamName = 'metamask-provider',
+      logger = console,
+      maxEventListeners = 100,
+    }: BaseProviderOptions = {},
```
rekmarks, line 124: ~Let's remove the default assignment to `{}` here since we're now expecting the consumers to always provide the appropriate arguments.~

src/BaseProvider.ts
```diff
@@ -0,0 +1,531 @@
...
+      this._state.isConnected = true;
+    });
+
+    // setup RPC connection
+
+    const jsonRpcConnection = createStreamMiddleware();
+    pump(
+      jsonRpcConnection.stream,
+      mux.createStream(jsonRpcStreamName) as unknown as Duplex,
+      // jsonRpcConnection.stream,
```
rekmarks: I don't believe that this should be commented.

src/BaseProvider.ts
```diff
@@ -0,0 +1,531 @@
...
+      }
+
+      this._state.accounts = _accounts as string[];
+
+      // handle selectedAddress
+      if (this.selectedAddress !== _accounts[0]) {
+        this.selectedAddress = _accounts[0] as string || null;
+      }
+
+      // finally, after all state has be/.senen updated, emit the event
```
rekmarks:
```suggestion
      // finally, after all state has been updated, emit the event
```

src/MetaMaskInpageProvider.ts
```diff
@@ -149,144 +88,42 @@ export default class MetaMaskInpageProvider extends SafeEventEmitter {
...
-
-    // EIP-1193 connect
-    this.on('connect', () => {
-      this._state.isConnected = true;
-    });
+    // mux.ignoreStream('phishing');
 
     // setup RPC connection
 
     const jsonRpcConnection = createStreamMiddleware();
```
rekmarks: Rather than calling `createStreamMiddleware` in the constructor of this class in addition to calling it in `BaseProvider`, we should make the `jsonRpcConnection` created in `BaseProvider` accessible to subclasses via a `protected` property.

src/MetaMaskInpageProvider.ts
```diff
@@ -1,136 +1,75 @@
...
 }
 
-export interface SendSyncJsonRpcRequest extends JsonRpcRequest<unknown> {
-  method: 'eth_accounts' | 'eth_coinbase' | 'eth_uninstallFilter' | 'net_version';
-}
+export default class MetaMaskInpageProvider extends BaseProvider {
 
-interface InternalState {
-  sentWarnings: {
+  protected sentWarnings: SentWarningsState = {
```
rekmarks: Since this property will be accessible to consumers, let's prefix it to encourage people not to mess with it:

```suggestion
  protected _sentWarnings: SentWarningsState = {
```

src/MetaMaskInpageProvider.ts
rekmarks: Huh, sure, why not!?

I take it we can accomplish the same thing with optional chaining?

```suggestion
    if (this?.sentWarnings.events[eventName as WarningEventName] === false) {
```

rekmarks (changes requested):
Looking good! Just a few more things to iron out.

src/BaseProvider.ts
shanejonas: i get some issues when i remove it:

```sh
● BaseProvider: RPC › ._rpcRequest › calls _handleAccountsChanged on request for eth_accounts

    TypeError: Cannot read property 'jsonRpcStreamName' of undefined

      110 |     connectionStream: typeof Duplex,
      111 |     {
    > 112 |       jsonRpcStreamName = 'metamask-provider',
          |       ^
      113 |       logger = console,
      114 |       maxEventListeners = 100,
      115 |     }: BaseProviderOptions,

      at new BaseProvider (src/BaseProvider.ts:112:7)
      at initializeProvider (test/BaseProvider.rpc.test.js:11:20)
      at Object.<anonymous> (test/BaseProvider.rpc.test.js:204:18)
      ```

Comment (rekmarks) [reply src/BaseProvider.ts:124]: Oh, my bad. I mean to delete this suggestion after realizing it would probably cause issues 🤦 

Please ignore this suggestion!

File: src/MetaMaskInpageProvider.ts [hunk 17]
```diff
@@ -182,30 +181,6 @@ export default class MetaMaskInpageProvider extends BaseProvider {
...
-    if (
-      this._state.isConnected ||
-      (!this._state.isPermanentlyDisconnected && !isRecoverable)
-    ) {
-
-      if (!isRecoverable) {
-        this.networkVersion = null;
-      }
-    }
-  }
```
rekmarks: I think this was correct. We should set `this.networkVersion` to `null` if the disconnection is non-recoverable.

_Note:_ I accidentally made this comment on a removed commit but it reflects the current state of the PR.

rekmarks (changes requested):

src/MetaMaskInpageProvider.ts
rekmarks: I suppose it could be simplified to this:

```suggestion
  protected _handleDisconnect(isRecoverable: boolean, errorMessage?: string) {
    super._handleDisconnect(isRecoverable, errorMessage);
    if (this.networkVersion && !isRecoverable) {
      this.networkVersion = null;
    }
  }
```

test/BaseProvider.rpc.test.js
```diff
@@ -0,0 +1,307 @@
+const { BaseProvider } = require('../dist');
```
rekmarks, line 1: For posterity, the reason these tests worked with the `BaseProvider` streams set up incorrectly (see comment and changes [here](https://github.com/MetaMask/inpage-provider/pull/144#discussion_r622299174)) is because these tests mock `JsonRpcEngine.handle`, and so we're not testing the stream piping underlying that method at all.

rekmarks (approved):
LGTM!

PR #146, opened 2021-04-29
fix: add metamask eslint configs

shanejonas:
only things manually touched were package.json scripts, and added a couple ignore statements


rekmarks (approved):
LGTM!

PR #152, opened 2021-04-30
feat: add createExternalExtensionProvider

shanejonas:
- remove index.d.ts


src/createMetaMaskExternalExtensionProvider.ts
```diff
@@ -0,0 +1,34 @@
...
+
+const browser = detect();
+
+export default function createMetaMaskExternalExtensionProvider() {
+  let provider;
+  try {
+    const currentMetaMaskId = getMetaMaskId();
+    const metamaskPort = chrome.runtime.connect(
+      currentMetaMaskId,
+    ) as Runtime.Port;
```
shanejonas: I have to cast this to play nicely with the `webextension-polyfill-ts` type that `extension-port-stream` uses, doesnt seem to match exactly to the type `chrome.runtime.connect` returns which is a `chrome.runtime.Port`

rekmarks (changes requested):
Starting to look at this now. We should delete `index.d.ts` in a separate PR.

src/createMetaMaskExternalExtensionProvider.ts
rekmarks: Makes sense. Port- and stream-related types seem to be incompatible 100% of the time.

rekmarks (changes requested):
Looks good! I have one request and one question:

Request: Since there are multiple of them, can we move all of the extension-related files into a new folder `src/extension-provider`?

Question: I'd like us to only use the `BaseProvider` instead of the inpage provider, to force consumers of this off the legacy API. What do you think of that?

shanejonas:
index.d.ts removal in this PR: https://github.com/MetaMask/inpage-provider/pull/153

shanejonas:
added `src/extension-provider/`

rekmarks (approved):
LGTM!

PR #167, opened 2021-05-05
Migrate tests to TypeScript

### LINKED ISSUE #147: migrate tests to typescript
currently they are in javascript and require stuff out of `../dist`

shanejonas:
fixes #147


jest.config.js
```diff
@@ -11,9 +11,9 @@ module.exports = {
   //   },
   // },
   setupFilesAfterEnv: ['./jest.setup.js'],
-  moduleFileExtensions: ['js'],
-  silent: true,
   testEnvironment: 'jsdom',
-  testRegex: ['\\.test\\.js$'],
+  silent: true,
```
rekmarks, line 15: Huh, TIL this flag is (allegedly) only available for the CLI, not config files:

- https://jestjs.io/docs/configuration
- https://jestjs.io/docs/cli

Does the behavior change at all when you toggle this flag? If not, we should remove it and pass `--silent` as part of the `test` script in `package.json`.

src/mocks/DuplexStream.ts
```diff
@@ -0,0 +1,21 @@
+import { Duplex } from 'stream';
```
rekmarks: I think we should keep mocks out of the `src` directory. Maybe recreate the `test` directory and save this as `test/MockDuplexStream.ts`?

AFAICT, it's currently being compiled by `yarn build`, but it's either way easier to prevent that from happening by keeping it in a separate directory altogether.

rekmarks (changes requested):
Praise be to Hejlsberg for `as any`!

Just a couple of requests.

rekmarks (approved):
LGTM!

PR #171, opened 2021-05-11
fix: _handleChainChanged in MetaMaskInpageProvider to handle networkVersion

shanejonas:
- adds tests for the 2 events `chainChanged` and `networkChanged`

### DIFF
--- a/src/BaseProvider.rpc.test.ts
+++ b/src/BaseProvider.rpc.test.ts
@@ -309,4 +309,24 @@ describe('BaseProvider: RPC', () => {
       });
     });
   });
+  describe('provider events', () => {
+    // eslint-disable-next-line jest/no-test-callback
+    it('calls chainChanged when it chainId changes ', (done) => {
+      const mockStream = new MockDuplexStream();
+      const p = new BaseProvider(mockStream);
+      (p as any)._state.initialized = true;
+      p.on('chainChanged', (changed) => {
+        expect(changed).toBeDefined();
+        done();
+      });
+      mockStream.push({
+        name: 'metamask-provider',
+        data: {
+          jsonrpc: '2.0',
+          method: 'metamask_chainChanged',
+          params: { chainId: '0x1' },
+        },
+      });
+    });
+  });
 });
--- a/src/BaseProvider.rpc.test.ts
+++ b/src/BaseProvider.rpc.test.ts
@@ -310,22 +310,23 @@ describe('BaseProvider: RPC', () => {
     });
   });
   describe('provider events', () => {
-    // eslint-disable-next-line jest/no-test-callback
-    it('calls chainChanged when it chainId changes ', (done) => {
+    it('calls chainChanged when it chainId changes ', async () => {
       const mockStream = new MockDuplexStream();
-      const p = new BaseProvider(mockStream);
-      (p as any)._state.initialized = true;
-      p.on('chainChanged', (changed) => {
-        expect(changed).toBeDefined();
-        done();
-      });
-      mockStream.push({
-        name: 'metamask-provider',
-        data: {
-          jsonrpc: '2.0',
-          method: 'metamask_chainChanged',
-          params: { chainId: '0x1' },
-        },
+      const baseProvider = new BaseProvider(mockStream);
+      (baseProvider as any)._state.initialized = true;
+      await new Promise((resolve) => {
+        baseProvider.on('chainChanged', (changed) => {
+          expect(changed).toBeDefined();
+          resolve(undefined);
+        });
+        mockStream.push({
+          name: 'metamask-provider',
+          data: {
+            jsonrpc: '2.0',
+            method: 'metamask_chainChanged',
+            params: { chainId: '0x1', networkVersion: '0x1' },
+          },
+        });
       });
     });
   });
--- a/src/BaseProvider.ts
+++ b/src/BaseProvider.ts
@@ -420,21 +420,32 @@ export default class BaseProvider extends SafeEventEmitter {
    */
   protected _handleChainChanged({
     chainId,
+    networkVersion,
   }: { chainId?: string; networkVersion?: string } = {}) {
-    if (!chainId || typeof chainId !== 'string' || !chainId.startsWith('0x')) {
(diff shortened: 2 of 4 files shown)


src/MetaMaskInpageProvider.rpc.test.ts
```diff
@@ -636,4 +636,42 @@ describe('MetaMaskInpageProvider: RPC', () => {
...
+  describe('provider events', () => {
+    // eslint-disable-next-line jest/no-test-callback
+    it('calls chainChanged when it chainId changes ', (done) => {
+      const mockStream = new MockDuplexStream();
+      const p = new MetaMaskInpageProvider(mockStream);
+      (p as any)._state.initialized = true;
+      p.on('chainChanged', (changed) => {
+        expect(changed).toBe('0x1');
+        done();
+      });
```
rekmarks: We should avoid disabling `jest/no-test-callback` by removing the callback and wrapping these lines in a Promise and awaiting it at the end of the test. Similarly below.

src/MetaMaskInpageProvider.rpc.test.ts
```diff
@@ -636,4 +636,42 @@ describe('MetaMaskInpageProvider: RPC', () => {
       );
     });
   });
+  describe('provider events', () => {
```
rekmarks, line 640: nit: It'd be great if we could add some breathing space between `describe` and `it` blocks for those of us that navigate our codebase using `vim` 🙏 .

It's also more readable!

src/MetaMaskInpageProvider.ts
```diff
@@ -395,4 +395,54 @@ export default class MetaMaskInpageProvider extends BaseProvider {
...
     );
   }
+
+  /**
+   * Upon receipt of a new chainId and networkVersion, emits corresponding
+   * events and sets relevant public state.
+   * Does nothing if neither the chainId nor the networkVersion are different
+   * from existing values.
+   *
+   * @emits MetamaskInpageProvider#chainChanged
```
rekmarks, line 405: We should add `networkChanged` here as well:
```suggestion
   * @emits MetamaskInpageProvider#chainChanged
   * @emits MetamaskInpageProvider#networkChanged
```

src/BaseProvider.ts
```diff
@@ -420,32 +420,21 @@ export default class BaseProvider extends SafeEventEmitter {
...
-      this._handleDisconnect(true);
-    } else {
-      this._handleConnect(chainId);
+    this._handleConnect(chainId);
 
-      if (chainId !== this.chainId) {
-        this.chainId = chainId;
-        if (this._state.initialized) {
-          this.emit('chainChanged', this.chainId);
-        }
```
rekmarks: I think we actually want to keep all of this in the base provider implementation, because we do (unfortunately) set `networkVersion` to `loading` when the extension is unable to connect to the RPC endpoint of the current chain. We don't have another way of communicating this at the moment, although we probably should.

src/MetaMaskInpageProvider.ts
```diff
@@ -395,4 +395,54 @@ export default class MetaMaskInpageProvider extends BaseProvider {
...
+          this.emit('chainChanged', this.chainId);
+        }
+      }
+
+      if (networkVersion !== this.networkVersion) {
+        this.networkVersion = networkVersion;
+        if (this._state.initialized) {
+          this.emit('networkChanged', this.networkVersion);
+        }
+      }
```
rekmarks: After restoring the deleted lines from the `BaseProvider` implementation of this method, I think it would be possible to just call `super._handleChainChanged` and then just add these lines to this implementation, with one modification:

```javascript
      if (networkVersion !== 'loading' && networkVersion !== this.networkVersion) {
```

src/MetaMaskInpageProvider.rpc.test.ts
```diff
@@ -636,4 +636,42 @@ describe('MetaMaskInpageProvider: RPC', () => {
...
+    // eslint-disable-next-line jest/no-test-callback
+    it('calls chainChanged when it chainId changes ', (done) => {
+      const mockStream = new MockDuplexStream();
+      const p = new MetaMaskInpageProvider(mockStream);
```
rekmarks: nit: Single-character variable names serve the devil's workshop.

```suggestion
      const provider = new MetaMaskInpageProvider(mockStream);
```

src/BaseProvider.rpc.test.ts
```diff
@@ -309,4 +309,24 @@ describe('BaseProvider: RPC', () => {
       });
     });
   });
+  describe('provider events', () => {
```
rekmarks, line 312: See comments in the other test file for feedback that applies here as well.

rekmarks (changes requested):
This looks like what we need. Great addition with the tests, as always!

src/MetaMaskInpageProvider.rpc.test.ts
shanejonas: `provider` is a variable name in the above scope

src/MetaMaskInpageProvider.rpc.test.ts
shanejonas: prettier yells at me for this:

image

src/MetaMaskInpageProvider.rpc.test.ts
shanejonas: changed to `inpageProvider`

src/MetaMaskInpageProvider.rpc.test.ts
rekmarks: Ah! Yeah, my request was a little unclear. Prettier is unhappy with the newline after an opening bracket, parenthesis, or curly brace, but it's fine with the empty newlines in here, which is what I mean:

```javascript
describe('foo', () => {
  it('bar', () => {
    // ...
  })

  it('bar', () => {
    // ...
  })
})

describe('foo', () => {
  it('bar', () => {
    // ...
  })

  it('bar', () => {
    // ...
  })
})
```

rekmarks (approved):
LGTM!

PR #173, opened 2021-05-13
Fix 8.1.1 changelog, add publishConfig to package.json

rekmarks:
We should probably note in the changelog that the package is renamed 🤪 

Since the package is scoped, we should also really have the appropriate `publishConfig` in the manifest. Also fixes some package.json scripts.

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,7 +8,8 @@
     "node": ">=12.0.0"
   },
   "publishConfig": {
-    "access": "public"
+    "access": "public",
+    "registry": "https://registry.npmjs.org/"
   },
   "scripts": {
     "build": "mkdir -p dist && rm -rf dist/* && tsc --project .",
--- a/package.json
+++ b/package.json
@@ -7,6 +7,9 @@
   "engines": {
     "node": ">=12.0.0"
   },
+  "publishConfig": {
+    "access": "public"
+  },
   "scripts": {
     "build": "tsc --project .",
     "test": "jest",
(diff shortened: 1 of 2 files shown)


shanejonas (dismissed):

package.json
```diff
@@ -7,16 +7,18 @@
   "engines": {
     "node": ">=12.0.0"
   },
+  "publishConfig": {
+    "access": "public"
```
Gudahtt: This should list the registry as well. At least that's what we do everywhere else

```suggestion
    "access": "public",
    "registry": "https://registry.npmjs.org/"
```

shanejonas (approved):

package.json
```diff
@@ -7,16 +7,19 @@
...
-    "build": "tsc --project .",
+    "build": "mkdir -p dist && rm -rf dist/* && tsc --project .",
     "test": "jest",
     "coverage": "jest --coverage",
     "lint:eslint": "eslint . --cache --ext js,ts",
     "lint:misc": "prettier '**/*.json' '**/*.md' '**/*.yml' --single-quote --ignore-path .gitignore",
     "lint": "yarn lint:eslint && yarn lint:misc --check",
     "lint:fix": "yarn lint:eslint --fix && yarn lint:misc --write",
-    "prepublishOnly": "yarn require-clean-git && yarn lint && yarn test",
-    "require-clean-git": "git diff --quiet || (echo 'Please clean the working directory.' && exit 1)"
```
rekmarks, line 19: For posterity: This is not a bad thing to do, but we don't do it anywhere else, and we probably aren't going to start.

PR #174, opened 2021-05-13
Move @types/chrome to dev dependencies

rekmarks:
Type dependencies should always be in dev deps.


shanejonas (approved):

Gudahtt:
I was pretty sure they needed to be included in the production dependencies for consumers to use the types. At least if they're a part of the API anyway.

rekmarks:
Oh shoot, I might be misremembering what we want to do here. Moving back to draft.

rekmarks:
I misremembered what we wanted to do here.

PR #178, opened 2021-06-22
Add @lavamoat/allow-scripts

rekmarks:
Adds `@lavamoat/allow-scripts` and related scripts / CI steps.


ryanml (approved):
++

Gudahtt:
We should follow up with adding a contributor section to the README. I usually do it when adding `allow-scripts` because the `setup` script is pretty unexpected for new contributors.

PR #180, opened 2021-06-22
Fix repo standardization issues, update Jest config

rekmarks:
- Add contributor instructions to readme (thanks @Gudahtt)
- Update lint dependencies
  - Some lint deps were complaining about unsupported versions
- Update Jest config; update tests to support updated config
  - All fake timers are removed. IIRC we used to have timers in certain parts of the source code, but we don't any longer. After updating the Jest configs, the fake timers appeared to cause some tests to fail.
  - This includes adopting our standard Jest config, and using the `projects` option to specify different test environments for the `BaseProvider` and the browser environment providers.


shanejonas (approved):

Gudahtt (approved):
LGTM! These project regexes seem like they'll be a pain to maintain as we add more tests. But we can deal with that later.

PR #194, opened 2021-09-11
feat: add unlockStateChanged event for dapps

rynst:
https://github.com/MetaMask/metamask-extension/issues/10368

Emit unlockStateChanged event to dapps with the status of whether or not it isUnlocked.  

A use case for this event could be a dapp that needs to know whether or not the user has triggered 'lock' on metamask, and based on that event, would lock the user out of the dapp, prompting for re-login.

### DIFF
--- a/src/BaseProvider.ts
+++ b/src/BaseProvider.ts
@@ -511,7 +511,6 @@ export default class BaseProvider extends SafeEventEmitter {
    * array.
    *
    * Does nothing if the received value is equal to the existing value.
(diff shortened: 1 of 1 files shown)


rsteubs:
Closing this PR since the [issue](https://github.com/MetaMask/metamask-extension/issues/10368) is resolved.

rynst:
Closing this PR since the [issue](https://github.com/MetaMask/metamask-extension/issues/10368) is resolved.

PR #209, opened 2022-06-05
Move stream functionality from BaseProvider to new StreamProvider

rekmarks:
The constructor of the `BaseProvider` expected a duplex stream, and the class as a whole contained a considerable amount of logic specific to streams and stream handling. This PR extracts this and other dependent functionality into a new class, `StreamProvider`, which is now extended by `MetaMaskInpageProvider`, while `BaseProvider` is made abstract. This will be useful for replacing internal legacy providers (which do not always / ever use streams) with EIP-1193 equivalents throughout the MetaMask codebase.

This PR should be completely non-breaking except for the changes made to the `BaseProvider` class. In addition, `createExternalExtensionProvider`, `MetaMaskInpageProvider`, and `initializeProvider` should have no behavior changes at all.

In detail, this PR:
- Converts `BaseProvider` to an abstract class, and moves all stream-specific functionality into a new class, `StreamProvider`
- Moves `BaseProvider` tests to the `StreamProvider` tests, adds an initialization test case
- Adds optional `rpcMiddleware` parameter to `BaseProvider`
  - Middlewares previously instantiated and pushed inside of `BaseProvider` are now instead passed as constructor parameters to `StreamProvider` and `MetaMaskInpageProvider` as necessary.
- Moves `networkVersion`-related logic from the `BaseProvider` to `MetaMaskInpageProvider`, while continuing to rely on the `networkVersion` having the value of `'loading'` to detect when we are mid-network change in the `StreamProvider` (and by extension its child classes).
- Updates the Jest config, adds `jest-it-up` and minimum coverage requirements

### DIFF
--- a/.github/workflows/build-test-lint.yml
+++ b/.github/workflows/build-test-lint.yml
@@ -21,8 +21,22 @@ jobs:
       - run: yarn --frozen-lockfile
       - run: yarn allow-scripts
       - run: yarn build
+      - name: Require clean working directory after building
+        shell: bash
+        run: |
+          if ! git diff --exit-code; then
+            echo "Working tree dirty after building"
+            exit 1
+          fi
       - run: yarn lint
       - run: yarn test
+      - name: Require clean working directory after tests
+        shell: bash
+        run: |
+          if ! git diff --exit-code; then
+            echo "Working tree dirty after tests"
+            exit 1
+          fi
       - name: Validate RC changelog
         if: ${{ startsWith(github.ref, 'release/') }}
         run: yarn auto-changelog validate --rc
--- a/.github/workflows/build-test-lint.yml
+++ b/.github/workflows/build-test-lint.yml
@@ -21,28 +21,21 @@ jobs:
       - run: yarn --frozen-lockfile
       - run: yarn allow-scripts
       - run: yarn build
-      - name: Require clean working directory after building
-        shell: bash
-        run: |
-          if ! git diff --exit-code; then
-            echo "Working tree dirty after building"
-            exit 1
-          fi
       - run: yarn lint
       - run: yarn test
-      - name: Require clean working directory after tests
-        shell: bash
-        run: |
-          if ! git diff --exit-code; then
-            echo "Working tree dirty after tests"
-            exit 1
-          fi
       - name: Validate RC changelog
         if: ${{ startsWith(github.ref, 'release/') }}
         run: yarn auto-changelog validate --rc
       - name: Validate changelog
         if: ${{ !startsWith(github.ref, 'release/') }}
         run: yarn auto-changelog validate
+      - name: Require clean working directory
+        shell: bash
+        run: |
+          if ! git diff --exit-code; then
+            echo "Working tree dirty at end of job"
+            exit 1
+          fi
   all-jobs-pass:
     name: All jobs pass
     runs-on: ubuntu-20.04
--- a/jest.config.js
+++ b/jest.config.js
@@ -20,10 +20,10 @@ module.exports = {
   coveragePathIgnorePatterns: ['/node_modules/', '/mocks/', '/test/'],
   coverageThreshold: {
     global: {
-      branches: 51.35,
-      functions: 47.67,
-      lines: 55.05,
-      statements: 55.21,
+      branches: 55.61,
+      functions: 52.81,
+      lines: 58.33,
+      statements: 58.63,
     },
   },
   projects: [
@@ -31,7 +31,7 @@ module.exports = {
       ...baseConfig,
       displayName: 'StreamProvider',
       testEnvironment: 'node',
-      testMatch: ['**/StreamProvider.test.ts'],
+      testMatch: ['**/StreamProvider.test.ts', '**/utils.test.ts'],
     },
     {
       ...baseConfig,
--- a/jest.config.js
+++ b/jest.config.js
@@ -15,9 +15,9 @@ module.exports = {
   projects: [
     {
       ...baseConfig,
-      displayName: 'BaseProvider',
+      displayName: 'StreamProvider',
       testEnvironment: 'node',
-      testMatch: ['**/BaseProvider.test.ts'],
+      testMatch: ['**/StreamProvider.test.ts'],
     },
     {
       ...baseConfig,
--- a/jest.config.js
+++ b/jest.config.js
@@ -8,7 +8,6 @@ const baseConfig = {
   // original implementations, between each test. It does not affect mocked
   // modules.
   restoreMocks: true,
-  testTimeout: 2500,
 };
 
 module.exports = {
@@ -21,10 +20,10 @@ module.exports = {
   coveragePathIgnorePatterns: ['/node_modules/', '/mocks/', '/test/'],
   coverageThreshold: {
     global: {
-      branches: 43.11,
-      functions: 45.12,
-      lines: 47.09,
-      statements: 47.41,
+      branches: 50.22,
+      functions: 48.78,
+      lines: 53.44,
+      statements: 53.63,
     },
   },
   projects: [
@@ -46,4 +45,5 @@ module.exports = {
     },
   ],
   silent: true,
+  testTimeout: 2500,
 };
--- a/jest.config.js
+++ b/jest.config.js
@@ -12,6 +12,21 @@ const baseConfig = {
 };
 
 module.exports = {
+  collectCoverage: true,
+  collectCoverageFrom: [
+    '<rootDir>/**/src/**/*.ts',
+    '!<rootDir>/**/src/**/*.test.ts',
+  ],
+  coverageReporters: ['clover', 'json', 'lcov', 'text', 'json-summary'],
+  coveragePathIgnorePatterns: ['/node_modules/', '/mocks/', '/test/'],
+  coverageThreshold: {
+    global: {
+      branches: 43.11,
+      functions: 45.12,
+      lines: 47.09,
+      statements: 47.41,
+    },
+  },
   projects: [
     {
       ...baseConfig,
@@ -30,18 +45,5 @@ module.exports = {
       setupFilesAfterEnv: ['./jest.setup.js'],
     },
   ],
-  collectCoverage: true,
-  collectCoverageFrom: ['./src/**.ts'],
-  coverageReporters: ['text', 'html'],
-  coveragePathIgnorePatterns: ['/node_modules/', '/mocks/'],
-  // TODO: Require coverage when we're closer to home.
-  // coverageThreshold: {
-  //   global: {
-  //     branches: 100,
-  //     functions: 100,
-  //     lines: 100,
-  //     statements: 100,
-  //   },
-  // },
   silent: true,
 };
--- a/jest.config.js
+++ b/jest.config.js
@@ -20,10 +20,10 @@ module.exports = {
   coveragePathIgnorePatterns: ['/node_modules/', '/mocks/', '/test/'],
   coverageThreshold: {
     global: {
-      branches: 55.61,
-      functions: 52.81,
-      lines: 58.33,
-      statements: 58.63,
+      branches: 56.19,
+      functions: 53.33,
+      lines: 58.44,
+      statements: 58.73,
     },
   },
   projects: [
--- a/jest.config.js
+++ b/jest.config.js
@@ -16,14 +16,14 @@ module.exports = {
     '<rootDir>/**/src/**/*.ts',
     '!<rootDir>/**/src/**/*.test.ts',
   ],
-  coverageReporters: ['clover', 'json', 'lcov', 'text', 'json-summary'],
+  coverageReporters: ['html', 'json-summary', 'text'],
   coveragePathIgnorePatterns: ['/node_modules/', '/mocks/', '/test/'],
   coverageThreshold: {
     global: {
-      branches: 50.67,
-      functions: 50.63,
-      lines: 54.59,
-      statements: 54.76,
+      branches: 51.12,
+      functions: 48.24,
+      lines: 54.96,
+      statements: 55.12,
     },
   },
   projects: [
--- a/jest.config.js
+++ b/jest.config.js
@@ -20,10 +20,10 @@ module.exports = {
   coveragePathIgnorePatterns: ['/node_modules/', '/mocks/', '/test/'],
   coverageThreshold: {
     global: {
-      branches: 50.22,
-      functions: 48.78,
-      lines: 53.44,
-      statements: 53.63,
+      branches: 50.67,
+      functions: 50.63,
+      lines: 54.59,
+      statements: 54.76,
     },
   },
   projects: [
--- a/jest.config.js
+++ b/jest.config.js
@@ -20,10 +20,10 @@ module.exports = {
   coveragePathIgnorePatterns: ['/node_modules/', '/mocks/', '/test/'],
   coverageThreshold: {
     global: {
-      branches: 51.12,
-      functions: 48.24,
-      lines: 54.96,
-      statements: 55.12,
+      branches: 51.35,
+      functions: 47.67,
+      lines: 55.05,
+      statements: 55.21,
     },
   },
   projects: [
--- a/src/BaseProvider.ts
+++ b/src/BaseProvider.ts
@@ -10,7 +10,12 @@ import {
   JsonRpcMiddleware,
 } from 'json-rpc-engine';
 import messages from './messages';
-import { getRpcPromiseCallback, ConsoleLike, Maybe } from './utils';
+import {
+  getRpcPromiseCallback,
+  ConsoleLike,
+  Maybe,
+  isValidChainId,
+} from './utils';
 
 export interface UnvalidatedJsonRpcRequest {
   id?: JsonRpcId;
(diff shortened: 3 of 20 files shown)


src/StreamProvider.ts
```diff
@@ -0,0 +1,148 @@
...
+  private _handleStreamDisconnect(streamName: string, error: Error) {
+    let warningMsg = `MetaMask: Lost connection to "${streamName}".`;
+    if (error?.stack) {
+      warningMsg += `\n${error.stack}`;
+    }
+
+    this._log.warn(warningMsg);
+    if (this.listenerCount('error') > 0) {
+      this.emit('error', warningMsg);
+    }
```
rekmarks, line 148: I consolidated `logStreamDisconnectWarning` (previously defined in `src/utils.ts`) into this method.

src/StreamProvider.test.ts
```diff
@@ -0,0 +1,371 @@
...
+          expect(streamProvider.isConnected()).toBe(true);
+          resolve();
+        });
+      });
+
+      expect(requestMock).toHaveBeenCalledTimes(1);
+      expect(requestMock).toHaveBeenCalledWith({
+        method: 'metamask_getProviderState',
+      });
+    });
```
rekmarks: This is the only added test case, and this file should otherwise be equivalent to the deleted `BaseProvider.test.ts` file, absent whitespace.

src/extension-provider/createExternalExtensionProvider.ts
```diff
@@ -1,20 +1,26 @@
...
     const metamaskPort = chrome.runtime.connect(
       currentMetaMaskId,
     ) as Runtime.Port;
+
     const pluginStream = new PortStream(metamaskPort);
-    provider = new BaseProvider(pluginStream);
+    provider = new StreamProvider(pluginStream, {
+      logger: console,
+      rpcMiddleware: getDefaultExternalMiddleware(console),
+    });
```
rekmarks: This should preserve the existing functionality exactly.

src/initializeInpageProvider.ts
```diff
@@ -42,27 +42,27 @@ export function initializeProvider({
   shouldSetOnWindow = true,
   shouldShimWeb3 = false,
 }: InitializeProviderOptions): MetaMaskInpageProvider {
-  let provider = new MetaMaskInpageProvider(connectionStream, {
+  const provider = new MetaMaskInpageProvider(connectionStream, {
```
rekmarks, line 46: No need to use `let` here.

.github/workflows/build-test-lint.yml
```diff
@@ -21,8 +21,22 @@ jobs:
       - run: yarn --frozen-lockfile
       - run: yarn allow-scripts
       - run: yarn build
+      - name: Require clean working directory after building
```
mcmire: Worth adding this to the module template?

jest.config.js
```diff
@@ -8,16 +8,30 @@ const baseConfig = {
...
-  testTimeout: 2500,
 };
 
 module.exports = {
+  collectCoverage: true,
+  collectCoverageFrom: [
+    '<rootDir>/**/src/**/*.ts',
+    '!<rootDir>/**/src/**/*.test.ts',
+  ],
+  coverageReporters: ['clover', 'json', 'lcov', 'text', 'json-summary'],
```
mcmire: What does `clover` do? And also `lcov`? And `json-summary`?

src/BaseProvider.ts
```diff
@@ -66,21 +53,23 @@ export interface BaseProviderState {
...
+ * An abstract class implementing the EIP-1193 interface. Implementers must:
+ *
+ * 1. At initialization, push a middleware to the internal `_rpcEngine` that
+ *    hands off requests to the server and receives responses in return.
+ * 2. At initialization, retrieve initial state and call
+ *    {@link BaseProvider._initialize} **once**.
+ * 3. Ensure that the provider's state is synchronized with the wallet.
+ * 4. Ensure that notifications are received and emitted as appropriate.
+ */
+export default abstract class BaseProvider extends SafeEventEmitter {
```
mcmire: `abstract` 👏🏻

src/BaseProvider.ts
```diff
@@ -267,37 +202,38 @@ export default class BaseProvider extends SafeEventEmitter {
...
-          networkVersion: string;
-        };
+  protected _initialize(initialState?: {
+    accounts: string[];
+    chainId: string;
+    isUnlocked: boolean;
+    networkVersion: string;
+  }) {
+    if (this._state.initialized === true) {
+      throw new Error('Provider already initialized.');
```
mcmire, line 229: I like this pattern a lot 👏🏻

src/StreamProvider.ts
```diff
@@ -0,0 +1,149 @@
...
+   * @param options.jsonRpcStreamName - The name of the internal JSON-RPC stream.
+   * Default: metamask-provider
+   * @param options.logger - The logging API to use. Default: console
+   * @param options.maxEventListeners - The maximum number of event
+   * listeners. Default: 100
+   */
+  constructor(
+    connectionStream: Duplex,
+    {
+      jsonRpcStreamName = 'metamask-provider',
```
mcmire: Should this option have a default in case we or someone else wants to use it for another purpose? Maybe MetamaskInpageProvider should set this default?

src/StreamProvider.ts
```diff
@@ -0,0 +1,149 @@
...
+      );
+    }
+    this._initialize(initialState);
+  }
+
+  /**
+   * Called when connection is lost to critical streams. Emits an 'error' event
+   * from the provider with the error message and stack if present.
+   *
+   * @emits MetamaskInpageProvider#disconnect
```
mcmire: Should this be `BaseProvider#disconnect`?

src/StreamProvider.test.ts
```diff
@@ -0,0 +1,371 @@
...
+
+describe('StreamProvider', () => {
+  describe('constructor', () => {
+    it('initializes state and emits events', async () => {
+      const accounts = ['0xabc'];
+      const chainId = '0x1';
+      const networkVersion = '1';
+      const isUnlocked = true;
+
+      const requestMock = jest
```
mcmire, line 35: Instead of stubbing the prototype, is it possible to stub `streamProvider` since we have it below?

src/StreamProvider.ts
```diff
@@ -0,0 +1,149 @@
...
+      mux.createStream(jsonRpcStreamName) as unknown as Duplex,
+      this._jsonRpcConnection.stream,
+      this._handleStreamDisconnect.bind(this, 'MetaMask RpcProvider'),
+    );
+
+    // Wire up the JsonRpcEngine to the JSON-RPC connection stream
+    this._rpcEngine.push(this._jsonRpcConnection.middleware);
+
+    // Set initial state
+    this._initializeAsync();
```
mcmire: Hmm. I see that we were following this pattern before, but I don't think we should be doing this. I don't think people would expect a constructor to run code asynchronously and I feel like it would lead to weird behavior (and may also print warnings in tests as it is impossible to clean this up). 

In the past the way I've solved this problem is to move the async behavior to another method and have consumers call this themselves. So in this case, we could rename `_initializeAsync` to `initialize`, change it to a public method, and we would also (to reduce confusion) rename `_initialize` back to `_initializeState`. 

I get that the goal of this PR is not to change backward-compatible behavior, but what are your thoughts on making this change in a future PR?

mcmire (dismissed):
This is something that we've obviously been thinking about for a while and have wanted to do, so thank you for doing this. I took another look at EIP-1193 to verify that BaseProvider satisfies the minimum requirements and didn't see anything that stands out. I also like how minimal BaseProvider is now. A few questions/suggestions below, but overall looks good to me.

jest.config.js
rekmarks: These defaults are from somewhere, I believe @FrederikBolding knows.

Edit: These are the [jest defaults](https://jestjs.io/docs/configuration) plus `json-summary`, which is required by `jest-it-up`. @FrederikBolding, is there any reason to use the defaults instead of just `html` if we don't use them for anything? Our defaults are currently `html` and `text`.

rekmarks: [`lcov`](https://github.com/linux-test-project/lcov) is a GNU coverage tool. It also outputs an `index.html` file that looks identical to the one produced by the `html` reporter, it's just contained in a sub-directory named `lcov/`. I don't know how it's better/different. I also have no idea what `clover` does, other than output some files.

.github/workflows/build-test-lint.yml
rekmarks: We already have it after building, but we should probably add it after testing along with `jest-it-up`. See here: https://github.com/MetaMask/metamask-module-template/pull/106

src/BaseProvider.ts
rekmarks: It's too bad it doesn't appear to be reflect in the compiled JavaScript, but we do not appear to live in the best of all possible worlds. 😄

src/StreamProvider.ts
rekmarks: I agree with your assessment and suggestions. I made some steps in this direction 12ed69c and 33fc72c (the latter just renaming `_initialize` to `_initializeState`).

In essence, I kept `_initializeStateAsync` but made the original `StreamProvider` into an abstract class, `AbstractStreamProvider`. Neither calls `_initializeStateAsync` in its constructor, and the concrete class defines a public method `initialize()` for this purpose. The reason for this somewhat convoluted situation is that `MetaMaskInpageProvider` needs to inherit from the stream provider class, but we want to avoid changing that class to the greatest extent possible in this PR due to its sensitive nature.

So, `MetaMaskInpageProvider` now inherits from `AbstractStreamProvider`, and calls `_initializeStateAsync` in its constructor, preserving its existing behavior without modifying our usage of it anywhere. The good news is that the purpose of the inpage provider is to encapsulate horribleness, and I don't expect it to change before it's superseded by something better, so we can probably live with what I've done here.

src/StreamProvider.test.ts
rekmarks: I couldn't do that due to the async initialization stuff, but we now just stub the `streamProvider` directly.

src/StreamProvider.ts
rekmarks: Done in [315e2b2](https://github.com/MetaMask/providers/pull/209/commits/315e2b2699cb435c64d7b53682bf18559a24d6ab)

src/StreamProvider.ts
rekmarks: Done in [fc54527](https://github.com/MetaMask/providers/pull/209/commits/fc5452763e043eb888ce6b9b1a05bd1753efa24a)

jest.config.js
rekmarks: For now, I went with the same coverage reporters as in https://github.com/MetaMask/metamask-module-template/pull/106

FrederikBolding: @rekmarks @mcmire For skunkworks I used the `jest` defaults because we send those to CodeCov.io. If the repo doesn't use CodeCov.io then it's fine to use whatever as long as `json-summary` is included as it is required by `jest-it-up`.

It's worth noting though that the `jest` defaults also outputs the same HTML, so there isn't too much reason to customize the `coverageReporters` normally.

rekmarks: It looks like CodeCov only needs the `lcov` output, so we could probably get rid of everything else except `text` and `jest-it-up` if we were so inclined. I like `html` because it doesn't nest `index.html` inside of the `lcov/` directory, but 🤷‍♂️

FrederikBolding: CodeCov combines all the different formats and IIRC I tried with just one of the outputs which seemed to have less precision than when I used the default settings, so I just went with the defaults 😅

src/BaseProvider.ts
```diff
@@ -267,37 +202,40 @@ export default class BaseProvider extends SafeEventEmitter {
...
-      const { accounts, chainId, isUnlocked, networkVersion } =
-        (await this.request({
-          method: 'metamask_getProviderState',
-        })) as {
-          accounts: string[];
-          chainId: string;
-          isUnlocked: boolean;
-          networkVersion: string;
-        };
+  protected _initializeState(initialState?: {
```
mcmire, line 222: Sorry, one more thing 😅  This name implies that we are storing `accounts`, `chainId`, etc. in state, which isn't true. What if we rename this `_completeInitialization`? It's vague — `_handleNetworkChanges` or something might be more specific — but if we want to change this method later then we don't have to change the name.

mcmire (commented):
One more thing, then I'm good!

By the way one thing I've noticed is that you've changed all of the default exports to named exports. That isn't the first time I've seen that pattern, so I'm curious — what are your thoughts? What are the benefits you see using named exports vs. default exports? Are there any scenarios in which we would still want to use default exports?

src/StreamProvider.ts
mcmire: I understand the difficulty and I think your changes are a good compromise!

jest.config.js
mcmire: Cool, thanks for the insight guys!

.github/workflows/build-test-lint.yml
mcmire: Cool, thanks!

src/BaseProvider.ts
rekmarks: Ah, we very much do store those things in state. `accounts` and `isUnlocked` are stored in `this._state`, while `chainId` and `networkVersion` are stored as public properties under those names. This did alert me to a problem which is that `BaseProvider` still has logic to handle `networkVersion`, which is not part of the EIP-1193 interface. I moved that logic into `MetaMaskInpageProvider` in abd90a1. You can observe that the tests I added pass on `main` via this demonstration PR: https://github.com/MetaMask/providers/pull/210

_Edit:_ The reason this may not appear to update state is because it passes the values to a bunch of internal handler functions which ultimately set the values in their appropriate places.

rekmarks:
@mcmire, regarding named vs default exports, I think default exports make sense when you have a module that exports a single thing for which the name isn't that important. As soon as you are exporting multiple things from the same file, it's probably time for named exports. Mixing default and named exports isn't really that harmful in TypeScript IMO, but there's not really any benefit either. In JavaScript, I've found the practice to be actively harmful and confusing, because it's easy to mess up your imports and/or exports when your editor doesn't tell you about it.

The reason I made the change in this PR is because I made a change that caused a file to export two things instead of one, and then I just went through and removed all default exports except one, which is an internal log / error message utility (`src/messages.ts`).

On second thought, default exports also have the weakness of being importable under different names, which can be very confusing in large code bases. For our single message utility it doesn't matter, but it's been a real pain in the ass to deal with in the extension historically.

mcmire:
Okay, thanks for the explanation, makes sense!

src/BaseProvider.ts
mcmire: Ohhh I see. Thanks for the clarification. I'll take a look at that commit.

src/MetaMaskInpageProvider.test.ts
```diff
@@ -618,49 +661,113 @@ describe('MetaMaskInpageProvider: RPC', () => {
...
+      const [inpageProvider, mockStream] = await getInitializedProvider();
+
+      // We check this mostly for the readability of this test.
+      expect(inpageProvider.isConnected()).toBe(true);
+      expect(inpageProvider.chainId).toBe('0x0');
+      expect(inpageProvider.networkVersion).toBe('0');
+
+      const emitSpy = jest.spyOn(inpageProvider, 'emit');
+
+      inpageProvider.once('chainChanged', (newChainId) => {
```
mcmire: Should this wrapped in a promise to ensure that this gets checked before the test ends? Something like:

``` javascript
const promiseForChainChanged = new Promise((resolve) => {
  inpageProvider.once('chainChanged', (newChainId) => {
    expect(newChainId).toBe('0x1');
     resolve();
  });
});
```

and then later, after `mockStream.push` has been called:

``` javascript
await promiseForChainChanged;
```

rekmarks: Ah! That listener is completely useless. We check that `chainChanged` is called via the `emitSpy` at the end of the test. I removed this listener in 5bba4db.

src/StreamProvider.ts
```diff
@@ -0,0 +1,167 @@
...
+    if (this.listenerCount('error') > 0) {
+      this.emit('error', warningMsg);
+    }
+
+    this._handleDisconnect(false, error ? error.message : undefined);
+  }
+}
+
+/**
+ * An EIP-1193 provider wired to a some duplex stream via a
```
mcmire: Typo: "wired to a some" -> "wired to some"?

src/StreamProvider.ts
```diff
@@ -0,0 +1,167 @@
...
+
+    try {
+      initialState = (await this.request({
+        method: 'metamask_getProviderState',
+      })) as Parameters<BaseProvider['_initializeState']>[0];
+    } catch (error) {
+      this._log.error(
+        'MetaMask: Failed to get initial state. Please report this bug.',
+        error,
+      );
```
mcmire, line 128: Should we return after this or would passing `undefined` to `_initializeState` work?

mcmire (commented):
Two more things!

src/StreamProvider.ts
rekmarks: Fixed here: b8e9916

src/StreamProvider.ts
rekmarks: It's good as-is, `_initializeState` is designed to handle being passed `undefined`.

mcmire: Cool 👍🏻

mcmire (dismissed):
LGTM!

mcmire (approved):
LGTM!

PR #214, opened 2022-06-10
Use `_handleConnect` for initial `connect` event

rekmarks:
We were basically re-implementing the logic of `_handleConnect` by setting up a listener in the `BaseProvider` constructor that we never tore down. This PR just uses the dedicated handler for the event like nature intended.


mcmire (approved):
Nice, makes sense!

PR #217, opened 2022-07-06
Make window.ethereum accessible via top window only by default

weizman:
As part of an internal discussion it was decided that it's best that `window.ethereum` API will be exported via top window only (by default).

The decision is derived from the fact that when an iframe to domain X inside a webpage under domain Y communicates with MM via `window.ethereum`, MM displays the transaction's origin as X (even though the dapp's origin is Y).

As a rule of thumb (security-wise) we believe that MM should respect requests only from the domain of the dapp itself, and not from other domains that are loaded inside the dapp.

In case truly needed, we will consider creating a tool for delegation of such capabilities between top frame and child frames within a dapp (please reach out if this affects you).

> We use MM (MetaMask) just as an example, can be any other extension wallet as well)

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -20,7 +20,7 @@ module.exports = {
   coveragePathIgnorePatterns: ['/node_modules/', '/mocks/', '/test/'],
   coverageThreshold: {
     global: {
-      branches: 55.75,
+      branches: 54.78,
       functions: 52.81,
       lines: 58.22,
       statements: 58.52,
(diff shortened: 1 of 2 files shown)


kumavis (approved):

kumavis:
this requires a **major version bump** due to breaking change in injection behavior

weizman:
After some internal discussion there seem to be examples for legitimate scenarios where subframe should initiate txs with mm (e.g. uniswap iframe service) and therefore this PR should be withdrawn and revisited.

PR #218, opened 2022-07-19
Add deprecation warning for encryption methods

Gudahtt:
A deprecation warning has been added for each of the two encryption-related methods. The deprecation warning has been implemented as middleware and included in both external providers (the inpage provider and the external extension provider).


Gudahtt:
~This depends upon #219~

mcmire (approved):
Makes sense to me.

PR #219, opened 2022-07-19
Improve unit tests

Gudahtt:
The mocks used in the external provider unit tests have been rewritten to allow sending and inspecting messages in either direction. They also no longer rely upon mocking out internal methods.

The `getInitializedInpageProvider` helper function in the inpage provider tests has been updated with method callbacks, to allow tests to setup replies for specific messages. This will be used in a future PR.

One additional test was added to ensure the `isMetaMask` property is set on the inpage provider.


src/extension-provider/createExternalExtensionProvider.test.ts
```diff
@@ -1,31 +1,27 @@
 import { StreamProvider } from '../StreamProvider';
+import { MockPort } from '../../test/mocks/MockPort';
 import { createExternalExtensionProvider } from './createExternalExtensionProvider';
 
 describe('createExternalExtensionProvider', () => {
-  beforeEach(() => {
+  it('can be called and not throw', () => {
     (global.chrome.runtime.connect as any).mockImplementation(() => {
```
Gudahtt, line 8: Technically I didn't need to move this from the `beforeEach` into the individual tests here, but it's needed in the next PR because it conflicts with some of the setup in later tests.

mcmire, line 8: What's mocking `global.chrome.runtime.connect`? Why do we expect this to be a Jest mock function at this point in time?

src/MetaMaskInpageProvider.test.ts
```diff
@@ -11,45 +13,92 @@ import messages from './messages';
...
+        setImmediate(() => callback(data));
+      }
+    }
+    onWrite(name, data);
+  });
 
-  // Relinquish control of the event loop to ensure that the mocked state is
-  // retrieved.
-  await new Promise<void>((resolve) => setTimeout(() => resolve(), 1));
+  const provider = new MetaMaskInpageProvider(connectionStream);
```
mcmire, line 100: Can this be also written as this?

``` typescript
  await new Promise<void>((resolve) => {
    provider.on('_initialized', resolve);
  });
```

src/MetaMaskInpageProvider.test.ts
```diff
@@ -11,45 +13,92 @@ import messages from './messages';
...
+  const { resolve: resolveInitialized, promise: waitForInitialization } =
+    getDeconstructedPromise();
+  provider.on('_initialized', resolveInitialized);
 
-  expect(requestMock).toHaveBeenCalledTimes(1); // Sanity check
-  requestMock.mockRestore(); // Get rid of the mock
+  await waitForInitialization;
 
-  return [inpageProvider, mockStream] as const;
+  return [provider, connectionStream, onWrite] as const;
```
mcmire: Nit: I think it would be more flexible if this returned an object rather than an array, but this is probably fine for now.

src/MetaMaskInpageProvider.test.ts
```diff
@@ -11,45 +13,92 @@ import messages from './messages';
...
+    callback: (data: JsonRpcRequest<unknown>) => void;
+  }[];
+} = {}) {
+  let isInitialized = false;
+  const onWrite = jest.fn();
+  const connectionStream = new MockConnectionStream((name, data) => {
+    if (!isInitialized) {
+      if (
+        name === 'metamask-provider' &&
+        data.method === 'metamask_getProviderState'
```
mcmire: I notice that with MockDuplexStream, we didn't handle checking for `metamask_getProviderState`. I think I might know the answer to this, but why do we have to do this now?

src/extension-provider/createExternalExtensionProvider.test.ts
Gudahtt: It's added by `jest-chrome`, which is run in `jest.setup.js`

Gudahtt: To be honest I didn't know the answer to this either, I was just moving the pre-existing call. I'll add a comment to clarify.

src/MetaMaskInpageProvider.test.ts
Gudahtt: Good question! Previously the initial provider state was set by mocking out the `request` function and returning the initial state directly, bypassing the entire RPC layer. `metamask_getProviderState` was never really called, and any code past that `request` call was not run at all.

That mock was removed, and instead now we're using a mock _connection_ instead. As such, now the _entire_ initialization process is being tested, rather than just up to the `.request` invocation.

This means our mock has to behave roughly like a real connection with the wallet would though, so it's a bit more complex. This mock is setup to handle the initialization call, then pass along all other calls to the configured callbacks.

src/MetaMaskInpageProvider.test.ts
Gudahtt: I agree. I left it as an array to minimize changes, but... I'll take another look, it might not be that messy to switch.

src/MetaMaskInpageProvider.test.ts
Gudahtt: Yes.... good point. The `getDeconstructedPromise` pattern was used _very heavily_ before I came up with this method callback idea. I hadn't come back to this to re-think it since then. Your suggested alternative is much more concise, I'll update this.

src/extension-provider/createExternalExtensionProvider.test.ts
mcmire: Ahh cool, that makes sense.

src/MetaMaskInpageProvider.test.ts
mcmire: Very cool. I like that!

src/extension-provider/createExternalExtensionProvider.test.ts
Gudahtt: Done here: ebea8b835713396d4b585754fd6e58825b9752dc

src/MetaMaskInpageProvider.test.ts
Gudahtt: Done here: 73745978b394740335ddb34f19ae854fcc43bf33

src/MetaMaskInpageProvider.test.ts
Gudahtt: Done here: 6508da238b31dabd661a0358781f3c8d2613e2a3

src/MetaMaskInpageProvider.test.ts
Gudahtt: This conversation had me re-thinking this initialization code, and in retrospect I think this protection is unnecessary and potentially harmful. This logic was simplified here, with the explanation for why in the commit message: 8eee8b88f93c95956ef0e8c94227cc2ffeea951e

mcmire: Ah okay. Sounds good!

mcmire (approved):
Nice. Looks good to me!

PR #223, opened 2022-09-23
Passing configuration options to JSON stream middleware

jpuri:
PR should be merged only after https://github.com/MetaMask/json-rpc-middleware-stream/pull/27 is deployed.

`json-rpc-middleware-stream` takes care to replay DAPP action when reconnection message is received. This PR passes the configuration parameter for the replay message in `json-rpc-middleware-stream`.


mcmire:
It looks like the new version of `json-rpc-middleware-stream` drops support for Node 12, upgrading the `engines` field to `>=14.0.0`. We would need to do the same thing for this library to be able to use this version.

jpuri:
PR is blocked by https://github.com/MetaMask/providers/pull/225

jpuri:
The PR is ready for review @mcmire , @Gudahtt

src/StreamProvider.ts
```diff
@@ -71,7 +71,9 @@ export abstract class AbstractStreamProvider extends BaseProvider {
     );
 
     // Set up RPC connection
-    this._jsonRpcConnection = createStreamMiddleware();
+    this._jsonRpcConnection = createStreamMiddleware({
+      retryOnMessage: 'METAMASK_EXTENSION_STREAM_CONNECT',
```
Gudahtt, line 75: Where is this message sent? Just trying to understand the whole solution.

jpuri: The message is sent from contentScript when extension stream is connected https://github.com/MetaMask/metamask-extension/blob/develop/app/scripts/contentscript.js#L395

Gudahtt (approved):
LGTM!

PR #228, opened 2022-10-28
Update json-rpc-middleware-stream

jpuri:
- update `json-rpc-middleware-stream`
- rename constant as discussed here https://github.com/MetaMask/metamask-extension/pull/16250#discussion_r1007779102


Gudahtt (approved):
LGTM!

PR #248, opened 2023-03-28
fix accountsChanged callback loop

Louis-kupo:
### Reason
Change the **accountsChanged** callback object to a new object , or will cause loop.This is because the callback function of accountsChanged returns the reference of the accounts, so request **eth_accounts** after externally modifying the reference of the account will cause accountsChanged to be triggered again, then it keeps looping
### Reproducibility
If you change the result returned by **accountsChanged** and request **eth_accounts** again, you will enter a loop .
```js
  window.ethereum.on("accountsChanged", async (newAccounts) => {
    newAccounts[0] = "1234...";
    const nextAccounts = await window.ethereum.request({
      method: "eth_accounts"
    });
    console.log("nextAccounts", nextAccounts);
  });
```
### Solution
In the callback of accountsChanged, return the deep copy object

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -22,8 +22,8 @@ module.exports = {
     global: {
       branches: 58.12,
       functions: 54.95,
-      lines: 60.76,
-      statements: 60.99,
+      lines: 60.61,
+      statements: 60.84,
     },
   },
   projects: [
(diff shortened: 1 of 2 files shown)


Gudahtt:
Thanks, great find! This seems like the right solution to me.

Could you fix the lint error? `yarn lint:fix` should fix the lint error being highlighted by CI.

Louis-kupo:
> Thanks, great find! This seems like the right solution to me.
> 
> Could you fix the lint error? `yarn lint:fix` should fix the lint error being highlighted by CI.

ok

Gudahtt (approved):
LGTM, thanks!

I pushed one additional commit to adjust the code coverage requirements to get CI to pass. While it would be ideal to have a unit test for this error, the file being changed is not yet tested.

PR #252, opened 2023-04-19
Support Flask and Beta in the external extension provider

### LINKED ISSUE #251: `createExternalExtensionProvider` does not support connecting to Flask
Since the connection IDs are hardcoded, Flask is not currently supported: https://github.com/MetaMask/providers/blob/a6c77a49f5b0b9fb192ebf1036c5f4aa6a5e0643/src/extension-provider/createExternalExtensionProvider.ts#L38-L47

FrederikBolding:
Adds support for Flask, Beta and custom extension IDs in the external extension provider.

Fixes #251


src/extension-provider/createExternalExtensionProvider.test.ts
```diff
@@ -126,6 +127,40 @@ describe('createExternalExtensionProvider', () => {
     expect(results).toBeInstanceOf(StreamProvider);
   });
 
+  it('supports Flask', () => {
+    // `global.chrome.runtime` mock setup by `jest-chrome` in `jest.setup.js`
+    (global.chrome.runtime.connect as any).mockImplementation(() => {
```
Mrtenz, line 132:
```suggestion
    (global.chrome.runtime.connect as MockedFunction<typeof global.chrome.runtime.connect>).mockImplementation(() => {
```

Same for the other tests.

FrederikBolding: If we don't cast to any it will complain about the `MockPort` class not matching with types. Seems out of scope imo

src/extension-provider/createExternalExtensionProvider.ts
```diff
@@ -35,13 +37,9 @@ export function createExternalExtensionProvider() {
...
-    case 'firefox':
-      return config.FIREFOX_ID;
-    default:
-      return config.CHROME_ID;
-  }
+function getExtensionId(typeOrId: ExtensionType) {
+  const ids =
+    browser?.name === 'firefox' ? config.firefoxIds : config.chromeIds;
+  // @ts-expect-error Ignoring indexing error since we have a fallback value
+  return ids[typeOrId] ?? typeOrId;
```
Mrtenz: Would prefer some kind of type cast here over ignoring TS errors for the entire line.
```suggestion
  return ids[typeOrId as keyof typeof ids] ?? typeOrId;
```

FrederikBolding: Good point!

src/extension-provider/createExternalExtensionProvider.ts
```diff
@@ -8,14 +8,16 @@ import config from './external-extension-config.json';
 
 const browser = detect();
 
-export function createExternalExtensionProvider() {
+export type ExtensionType = 'stable' | 'flask' | 'beta' | string;
```
Mrtenz, line 11: Does this actually provide autocomplete, since `ExtensionType` is inferred as just `string`?

Mrtenz (approved):

src/extension-provider/createExternalExtensionProvider.ts
FrederikBolding: It does not. Do you have an idea of how to make that happen?

Mrtenz: Probably not possible unless the ID is not a string.

```ts
export type ExtensionType = 'stable' | 'flask' | 'beta' | `id:${string}`;
```

This works for example, but not sure if that's what we want.

PR #254, opened 2023-04-24
Standardise repo per module template as of April 2023

Mrtenz:
This updates the repo to the latest standards. This includes a bump to the ESLint configurations, meaning I had to fix a bunch of ESLint-related errors.

## Breaking changes

- The minimum Node.js version is now 16.


src/extension-provider/createExternalExtensionProvider.test.ts
```diff
@@ -61,10 +58,10 @@ async function getInitializedProvider({
       name === 'metamask-provider' &&
       data.method === 'metamask_getProviderState'
     ) {
-      // Wrap in `setImmediate` to ensure a reply is recieved by the provider
+      // Wrap in `setTimeout` to ensure a reply is received by the provider
       // after the provider has processed the request, to ensure that the
       // provider recognizes the id.
-      setImmediate(() =>
+      setTimeout(() =>
```
Mrtenz, line 64: `setImmediate` is not available in recent versions of Node.js and/or browsers, so I've swapped this out to use `setTimeout`. For our purposes, it does effectively the same.

src/MetaMaskInpageProvider.ts
```diff
@@ -298,6 +308,7 @@ export class MetaMaskInpageProvider extends AbstractStreamProvider {
    */
   send<T>(payload: SendSyncJsonRpcRequest): JsonRpcResponse<T>;
 
+  // eslint-disable-next-line @typescript-eslint/promise-function-async
```
Mrtenz, line 311: `send` can return a `Promise`, so ESLint wants to add `async` to it. This changes the behaviour of the function, however, as anything would make it return a `Promise`. To avoid this we have to suppress this rule here.

tsconfig.json
```diff
@@ -1,16 +1,18 @@
...
+    "noErrorTruncation": true,
+    "noUncheckedIndexedAccess": true,
     "strict": true,
-    "target": "ES2017",
-    "typeRoots": ["./node_modules/@types"]
+    "target": "es2020",
+    "resolveJsonModule": true,
+    "skipLibCheck": true
   },
-  "include": ["./src/*.ts"],
```
FrederikBolding, line 14: Do we not need an `include`?

.github/workflows/publish-release.yml
```diff
@@ -1,29 +1,116 @@
...
+          path: |
+            ./dist
+            ./node_modules/.yarn-state.yml
+          key: ${{ github.sha }}
+      - name: Publish
+        uses: MetaMask/action-npm-publish@v2
+        with:
+          # This `NPM_TOKEN` needs to be manually set per-repository.
+          # Look in the repository settings under "Environments", and set this token in the `npm-publish` environment.
+          npm-token: ${{ secrets.NPM_TOKEN }}
```
FrederikBolding, line 79: We shouldn't merge this until all tokens have been added to this repo

src/extension-provider/createExternalExtensionProvider.test.ts
```diff
@@ -174,7 +171,7 @@ describe('createExternalExtensionProvider', () => {
     ];
 
     for (const { method, warning } of warnings) {
-      describe(method, () => {
+      describe(`${method}`, () => {
```
FrederikBolding, line 174: ?

src/siteMetadata.ts
```diff
@@ -83,7 +89,7 @@ async function getSiteIcon(
   const icons: NodeListOf<HTMLLinkElement> = document.querySelectorAll(
     'head > link[rel~="icon"]',
   );
-  for (const icon of icons) {
+  for (const icon of Array.from(icons)) {
```
FrederikBolding, line 92: Shouldn't this already be iterable?

package.json
```diff
@@ -2,48 +2,46 @@
...
+    "build:docs": "typedoc",
+    "lint": "yarn lint:eslint && yarn lint:misc --check && yarn lint:dependencies --check && yarn lint:changelog",
+    "lint:changelog": "auto-changelog validate",
+    "lint:dependencies": "depcheck && yarn dedupe",
+    "lint:eslint": "eslint . --cache --ext js,ts",
+    "lint:fix": "yarn lint:eslint --fix && yarn lint:misc --write && yarn lint:dependencies && yarn lint:changelog",
+    "lint:misc": "prettier '**/*.json' '**/*.md' '!CHANGELOG.md' '**/*.yml' '!.yarnrc.yml' --ignore-path .gitignore --no-error-on-unmatched-pattern",
+    "prepack": "./scripts/prepack.sh",
+    "prepublishOnly": "yarn build && yarn lint && yarn test",
+    "setup": "yarn install && yarn allow-scripts",
```
FrederikBolding:
```suggestion
```

package.json
```diff
@@ -2,48 +2,46 @@
...
+    "build:clean": "rimraf dist && yarn build",
+    "build:docs": "typedoc",
+    "lint": "yarn lint:eslint && yarn lint:misc --check && yarn lint:dependencies --check && yarn lint:changelog",
+    "lint:changelog": "auto-changelog validate",
+    "lint:dependencies": "depcheck && yarn dedupe",
+    "lint:eslint": "eslint . --cache --ext js,ts",
+    "lint:fix": "yarn lint:eslint --fix && yarn lint:misc --write && yarn lint:dependencies && yarn lint:changelog",
+    "lint:misc": "prettier '**/*.json' '**/*.md' '!CHANGELOG.md' '**/*.yml' '!.yarnrc.yml' --ignore-path .gitignore --no-error-on-unmatched-pattern",
+    "prepack": "./scripts/prepack.sh",
+    "prepublishOnly": "yarn build && yarn lint && yarn test",
```
FrederikBolding:
```suggestion
```
I don't think this one is used anymore either?

package.json
```diff
@@ -52,34 +50,49 @@
     "is-stream": "^2.0.0",
     "json-rpc-engine": "^6.1.0",
     "json-rpc-middleware-stream": "^4.2.1",
-    "pump": "^3.0.0"
+    "pump": "^3.0.0",
+    "webextension-polyfill": "^0.10.0"
```
FrederikBolding, line 52: Why was this added? 🤔

src/StreamProvider.ts
```diff
@@ -58,27 +62,27 @@ export abstract class AbstractStreamProvider extends BaseProvider {
       throw new Error(messages.errors.invalidDuplexStream());
     }
 
-    // Bind functions to prevent consumers from making unbound calls
-    this._handleStreamDisconnect = this._handleStreamDisconnect.bind(this);
```
FrederikBolding: Is this safe to remove because it is now hash private?

FrederikBolding:
Please document the breaking changes in this PR in the description.

src/MetaMaskInpageProvider.test.ts
```diff
@@ -851,6 +866,8 @@ describe('MetaMaskInpageProvider: RPC', () => {
 
             await provider.request({ method });
 
+            console.log('consoleWarnSpy.mock.calls', consoleWarnSpy.mock.calls);
```
FrederikBolding:
```suggestion
```

FrederikBolding (changes requested):

src/siteMetadata.ts
Mrtenz: Not according to TypeScript.

package.json
Mrtenz: We import `Runtime` from it, so without, ESLint would complain.

src/StreamProvider.ts
Mrtenz: It's not used anywhere where it's not bound, so I think it's safe to remove. Otherwise we can't use a hash private function, as they are read-only.

tsconfig.json
Mrtenz: Apparently not. It works like this, and is the same as the module template.

package.json
FrederikBolding: Oh but we only use the type which is why it wasn't added before. Hmm 🤔

src/siteMetadata.ts
FrederikBolding: Hmm, it is untested. So that is slightly worrying

package.json
FrederikBolding: Not sure if we should ignore the eslint error or not

FrederikBolding: I assume everything has been working until now

src/StreamProvider.ts
FrederikBolding: Note that it is a breaking change now that it cannot be accessed anymore

src/siteMetadata.ts
FrederikBolding: Could we test it and make sure it doesn't break any existing logic?

Mrtenz: `Array.from` just creates an array from the given value, so I doubt it would break anything? Testing in a browser locally, these two are equivalent:

```ts
const icons = document.querySelectorAll(...);
for (const icon of icons) { ... }
for (const icon of Array.from(icons)) { ... }
```

package.json
Mrtenz: I don't think there's a downside to adding it?

src/StreamProvider.ts
Mrtenz: It's a private field, so I'm not sure if we need to consider this breaking. It's a change to the internal API.

.github/workflows/publish-release.yml
Gudahtt: Tokens have been setup

package.json
```diff
@@ -52,34 +48,49 @@
...
+    "prettier-plugin-packagejson": "^2.3.0",
+    "rimraf": "^3.0.2",
+    "ts-jest": "^28.0.7",
+    "ts-node": "^10.7.0",
+    "typedoc": "^0.23.15",
+    "typescript": "~4.8.4"
+  },
+  "packageManager": "yarn@3.2.1",
+  "engines": {
+    "node": ">=14.0.0"
```
Gudahtt: It seems that nvmrc has been updated to v16 but this is still on v14

Gudahtt: The README still says v14 as well

Mrtenz: Good catch! At least the engines part would be solved by MetaMask/metamask-module-template#191. 😄

src/StreamProvider.ts
FrederikBolding: Hmm, a lot of effort has been put into making `MetaMaskInpageProvider` non-breaking. Legacy functions etc, perhaps we shouldn't be changing the internal API for that? 

I wonder if some dapp exists that use the internal API because it is currently exposed

FrederikBolding (approved):

PR #249, opened 2023-03-30
chore: fix deprecated webextension-polyfill-ts

zouhangwithsweet:
fix: #242 #224

PR #259, opened 2023-05-10
move @types/ packages from deps to devDeps

legobeat:
Originally introduced in:
- #152
- #249


mcmire (approved):
This makes sense to me. I don't think these types are necessary to use this package in a TypeScript project.

PR #263, opened 2023-05-23
Implement EIP-6963

rekmarks:
Implements [EIP-6963](https://eips.ethereum.org/EIPS/eip-6963) according to the current state of the specification. Adds two new top-level imports,  `eip6963AnnounceProvider` and `eip6963RequestProvider`, used by us and the dapp, respectively.

For reference, here are some `EIP6963ProviderInfo` values that would work for us:
```ts
export const MetaMaskEIP6963ProviderInfo: EIP6963ProviderInfo = {
  uuid: uuid(),
  name: 'MetaMask', // Update depending on the build, probably
  icon: 'https://raw.githubusercontent.com/MetaMask/brand-resources/cb6fd847f3a9cc5e231c749383c3898935e62eab/SVG/metamask-fox.svg', // TODO: Find a shorter URL
};
```

The purpose of the `uuid` field is to help dapps notice if multiple wallets are claiming the same `name`. It needs to be a valid v4 UUID. IMO we should create a UUID at the point where we call `announceProvider` (i.e. the content script for the extension). We could also send it from the background.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -34,14 +34,22 @@ module.exports = {
     },
 
     {
+      files: ['EIP6963.test.ts', 'jest.setup.browser.js'],
+      rules: {
+        // We're mixing Node and browser environments in these files.
+        'no-restricted-globals': 'off',
+      },
+    },
+
+    {
       files: ['jest.setup.browser.js'],
       env: { browser: true },
+      // This file contains copypasta and we're not going to bother fixing these.
       rules: {
         'jest/require-top-level-describe': 'off',
         'jsdoc/require-description': 'off',
         'jsdoc/require-param-description': 'off',
         'jsdoc/require-param-type': 'off',
-        'no-restricted-globals': 'off',
       },
     },
   ],
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -26,11 +26,23 @@ module.exports = {
     },
 
     {
-      files: ['*.test.ts', '*.test.js'],
+      files: ['*.test.ts', '*.test.js', 'jest.setup*.js'],
       extends: [
         '@metamask/eslint-config-jest',
         '@metamask/eslint-config-nodejs',
       ],
+    },
+
+    {
+      files: ['jest.setup.browser.js'],
+      env: { browser: true },
+      rules: {
+        'jest/require-top-level-describe': 'off',
+        'jsdoc/require-description': 'off',
+        'jsdoc/require-param-description': 'off',
+        'jsdoc/require-param-type': 'off',
+        'no-restricted-globals': 'off',
+      },
     },
   ],
 
--- a/jest.setup.browser.js
+++ b/jest.setup.browser.js
@@ -1,0 +1,88 @@
+Object.assign(global, require('jest-chrome'));
+
+// Clean up jsdom between tests.
+// Courtesy @jhildenbiddle
+// https://github.com/jestjs/jest/issues/1224#issuecomment-716075260
+const sideEffects = {
+  document: {
+    addEventListener: {
+      fn: document.addEventListener,
+      refs: [],
+    },
+    keys: Object.keys(document),
+  },
+  window: {
+    addEventListener: {
+      fn: window.addEventListener,
+      refs: [],
+    },
+    keys: Object.keys(window),
+  },
+};
+
+// Lifecycle Hooks
+// -----------------------------------------------------------------------------
+beforeAll(async () => {
+  // Spy addEventListener
+  ['document', 'window'].forEach((obj) => {
+    const { fn } = sideEffects[obj].addEventListener;
+    const { refs } = sideEffects[obj].addEventListener;
+
+    /**
+     *
+     * @param type
+     * @param listener
+     * @param options
+     */
+    function addEventListenerSpy(type, listener, options) {
+      // Store listener reference so it can be removed during reset
+      refs.push({ type, listener, options });
+      // Call original window.addEventListener
+      fn(type, listener, options);
+    }
+
+    // Add to default key array to prevent removal during reset
+    sideEffects[obj].keys.push('addEventListener');
+
+    // Replace addEventListener with mock
+    global[obj].addEventListener = addEventListenerSpy;
+  });
+});
+
+// Reset JSDOM. This attempts to remove side effects from tests, however it does
+// not reset all changes made to globals like the window and document
+// objects. Tests requiring a full JSDOM reset should be stored in separate
+// files, which is only way to do a complete JSDOM reset with Jest.
+beforeEach(async () => {
+  const rootElm = document.documentElement;
+
+  // Remove attributes on root element
+  [...rootElm.attributes].forEach((attr) => rootElm.removeAttribute(attr.name));
+
+  // Remove elements (faster than setting innerHTML)
+  while (rootElm.firstChild) {
+    rootElm.removeChild(rootElm.firstChild);
+  }
+
+  // Remove global listeners and keys
+  ['document', 'window'].forEach((obj) => {
+    const { refs } = sideEffects[obj].addEventListener;
+
+    // Listeners
+    while (refs.length) {
+      const { type, listener, options } = refs.pop();
+      global[obj].removeEventListener(type, listener, options);
+    }
+
+    // This breaks coverage.
+    // // Keys
+    // Object.keys(global[obj])
+    //   .filter((key) => !sideEffects[obj].keys.includes(key))
+    //   .forEach((key) => {
+    //     delete global[obj][key];
+    //   });
+  });
+
+  // Restore base elements
+  rootElm.innerHTML = '<head></head><body></body>';
+});
--- a/jest.setup.browser.js
+++ b/jest.setup.browser.js
@@ -25,8 +25,7 @@ const sideEffects = {
 beforeAll(async () => {
   // Spy addEventListener
   ['document', 'window'].forEach((obj) => {
-    const { fn } = sideEffects[obj].addEventListener;
-    const { refs } = sideEffects[obj].addEventListener;
+    const { fn, refs } = sideEffects[obj].addEventListener;
 
     /**
      *
--- a/src/EIP6963.test.ts
+++ b/src/EIP6963.test.ts
@@ -1,5 +1,3 @@
-/* eslint-disable no-restricted-globals */
-
 import {
   announceProvider,
   requestProvider,
--- a/src/EIP6963.test.ts
+++ b/src/EIP6963.test.ts
@@ -1,34 +1,27 @@
-import {
-  announceProvider,
-  requestProvider,
-  MetaMaskEIP6963ProviderInfo,
-} from './EIP6963';
+import { announceProvider, requestProvider } from './EIP6963';
+
+const getProviderInfo = () => ({
+  name: 'test',
+  icon: 'https://wallet.io/icon.svg',
+  walletId: 'testWalletId',
+  uuid: 'testUuid',
+});
 
 describe('EIP6963', () => {
-  describe('MetaMaskEIP6963ProviderInfo', () => {
-    it('has expected shape and values', () => {
-      expect(MetaMaskEIP6963ProviderInfo).toStrictEqual({
-        walletId: 'io.metamask',
-        uuid: expect.any(String),
-        name: 'MetaMask',
-        icon: 'https://raw.githubusercontent.com/MetaMask/brand-resources/cb6fd847f3a9cc5e231c749383c3898935e62eab/SVG/metamask-fox.svg',
-      });
-    });
-  });
-
   describe('announceProvider', () => {
     it('should announce a provider', () => {
-      const provider = { name: 'test' };
+      const provider: any = { name: 'test' };
+      const providerDetail = { info: getProviderInfo(), provider };
       const dispatchEvent = jest.spyOn(window, 'dispatchEvent');
       const addEventListener = jest.spyOn(window, 'addEventListener');
 
-      announceProvider(provider as any);
+      announceProvider(providerDetail);
 
       expect(dispatchEvent).toHaveBeenCalledTimes(1);
       expect(dispatchEvent).toHaveBeenCalledWith(
         new CustomEvent('eip6963:announceProvider', {
           detail: {
-            info: MetaMaskEIP6963ProviderInfo,
+            info: getProviderInfo(),
             provider,
           },
         }),
@@ -38,51 +31,19 @@ describe('EIP6963', () => {
         'eip6963:requestProvider',
         expect.any(Function),
       );
-    });
-
-    it('should not be affected by modifying MetaMaskEIP6963ProviderInfo', () => {
-      const provider = { name: 'test' };
-      const dispatchEvent = jest.spyOn(window, 'dispatchEvent');
-      const addEventListener = jest.spyOn(window, 'addEventListener');
-
-      const { walletId } = MetaMaskEIP6963ProviderInfo;
-      (MetaMaskEIP6963ProviderInfo as any).walletId = 'foo';
-      announceProvider(provider as any);
-
-      expect(dispatchEvent).toHaveBeenCalledTimes(1);
-      expect(dispatchEvent).toHaveBeenCalledWith(
-        new CustomEvent('eip6963:announceProvider', {
-          detail: {
-            info: { ...MetaMaskEIP6963ProviderInfo, walletId },
-            provider,
-          },
-        }),
-      );
-      expect((dispatchEvent as any).mock.calls[0][0].detail).toStrictEqual({
-        info: { ...MetaMaskEIP6963ProviderInfo, walletId },
-        provider,
-      });
-
-      expect(addEventListener).toHaveBeenCalledTimes(1);
-      expect(addEventListener).toHaveBeenCalledWith(
-        'eip6963:requestProvider',
-        expect.any(Function),
-      );
-
-      // Reset
-      (MetaMaskEIP6963ProviderInfo as any).walletId = walletId;
     });
   });
 
   describe('requestProvider', () => {
     it('should receive an announced provider (called before announceProvider)', async () => {
       const provider: any = { name: 'test' };
+      const providerDetail = { info: getProviderInfo(), provider };
       const handleProvider = jest.fn();
       const dispatchEvent = jest.spyOn(window, 'dispatchEvent');
       const addEventListener = jest.spyOn(window, 'addEventListener');
 
       requestProvider(handleProvider);
-      announceProvider(provider);
+      announceProvider(providerDetail);
       await delay();
 
       expect(dispatchEvent).toHaveBeenCalledTimes(2);
@@ -105,18 +66,19 @@ describe('EIP6963', () => {
 
       expect(handleProvider).toHaveBeenCalledTimes(1);
       expect(handleProvider).toHaveBeenCalledWith({
-        info: MetaMaskEIP6963ProviderInfo,
+        info: getProviderInfo(),
         provider,
       });
     });
 
     it('should receive an announced provider (called after announceProvider)', async () => {
       const provider: any = { name: 'test' };
+      const providerDetail = { info: getProviderInfo(), provider };
       const handleProvider = jest.fn();
       const dispatchEvent = jest.spyOn(window, 'dispatchEvent');
       const addEventListener = jest.spyOn(window, 'addEventListener');
 
-      announceProvider(provider);
+      announceProvider(providerDetail);
       requestProvider(handleProvider);
       await delay();
 
@@ -142,7 +104,7 @@ describe('EIP6963', () => {
 
       expect(handleProvider).toHaveBeenCalledTimes(1);
       expect(handleProvider).toHaveBeenCalledWith({
-        info: MetaMaskEIP6963ProviderInfo,
+        info: getProviderInfo(),
         provider,
       });
     });
--- a/src/EIP6963.test.ts
+++ b/src/EIP6963.test.ts
@@ -4,11 +4,23 @@ const getProviderInfo = () => ({
   name: 'test',
   icon: 'https://wallet.io/icon.svg',
   walletId: 'testWalletId',
-  uuid: 'testUuid',
+  uuid: '1449211e-5560-4235-9ab1-582cbe2b165f',
 });
 
 describe('EIP6963', () => {
   describe('announceProvider', () => {
+    it('throws if the UUID is invalid', () => {
+      ['foo', null, undefined, Symbol('bar')].forEach((invalidUuid) => {
+        const provider: any = { name: 'test' };
+        const providerDetail = { info: getProviderInfo(), provider };
+        providerDetail.info.uuid = invalidUuid as any;
+
+        expect(() => announceProvider(providerDetail)).toThrow(
+          new Error('Invalid `uuid` field. Must be a v4 UUID.'),
+        );
+      });
+    });
+
     it('should announce a provider', () => {
       const provider: any = { name: 'test' };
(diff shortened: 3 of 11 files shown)


src/EIP6963.ts
```diff
@@ -0,0 +1,156 @@
...
+  return (
+    typeof info.icon === 'string' &&
+    isValidUrl(info.icon) &&
+    typeof info.name === 'string' &&
+    Boolean(info.name) &&
+    typeof info.uuid === 'string' &&
+    UUID_V4_REGEX.test(info.uuid) &&
+    typeof info.walletId === 'string' &&
+    Boolean(info.walletId)
+  );
```
rekmarks, line 139: @Mrtenz hey I hear you like superstruct

src/EIP6963.ts
```diff
@@ -0,0 +1,156 @@
...
+}
```
rekmarks: Normally I think we'd use `superstruct` or something for this, but it's important to keep the bundle size as small as possible.

Mrtenz: `superstruct` is only 11.5 kB (or 3.4 kB gzipped), and `@metmask/providers` is 75.1 kB (or 21.1 kB gzipped). I don't think it makes a significant difference. 😅

src/EIP6963.ts
```diff
@@ -0,0 +1,149 @@
...
+export function requestProvider<HandlerReturnType>(
+  handleProvider: (providerDetail: EIP6963ProviderDetail) => HandlerReturnType,
+): void {
+  window.addEventListener(
+    EIP6963EventNames.Announce as any,
+    (event: EIP6963AnnounceProviderEvent) => {
+      if (
+        event.type === EIP6963EventNames.Announce &&
+        isObject(event.detail?.provider)
+      ) {
```
jiexi: > object MUST be frozen by calling Object.freeze() on the value of the detail property.

`Object.isFrozen(event.detail?.provider)`


EDIT: misread. the whole providerDetail object must be frozen

src/EIP6963.ts
```diff
@@ -0,0 +1,149 @@
...
+/**
+ * Validates an {@link EIP6963ProviderDetail} object.
+ *
+ * @param providerDetail - The {@link EIP6963ProviderDetail} to validate.
+ * @returns Whether the {@link EIP6963ProviderDetail} is valid.
+ */
+function isValidProviderDetail(
+  providerDetail: unknown,
+): providerDetail is EIP6963ProviderDetail {
+  if (!isObject(providerDetail) || !isObject(providerDetail.info)) {
```
jiexi: frozen check on `providerDetail.provider` here too?

EDIT: misread. the whole providerDetail object must be frozen

jest.setup.browser.js
```diff
@@ -0,0 +1,88 @@
...
+  },
+};
+
+// Lifecycle Hooks
+// -----------------------------------------------------------------------------
+beforeAll(async () => {
+  // Spy addEventListener
+  ['document', 'window'].forEach((obj) => {
+    const { fn } = sideEffects[obj].addEventListener;
+    const { refs } = sideEffects[obj].addEventListener;
```
jiexi: i think these can be combined `const { fn, refs } = sideEffects[obj].addEventListener;`

src/EIP6963.test.ts
```diff
@@ -0,0 +1,174 @@
...
+
+      // Notice that 3 events are dispatched in total when requestProvider is
+      // called after announceProvider.
+      expect(dispatchEvent).toHaveBeenCalledTimes(3);
+      expect(dispatchEvent).toHaveBeenCalledWith(
+        new Event('eip6963:requestProvider'),
+      );
+      expect(dispatchEvent).toHaveBeenCalledWith(
+        new CustomEvent('eip6963:announceProvider'),
+      );
```
jiexi: should we explicitly check that the first call is a announceProvider?

src/EIP6963.test.ts
```diff
@@ -0,0 +1,174 @@
...
+      announceProvider(providerDetail);
+      await delay();
+
+      expect(dispatchEvent).toHaveBeenCalledTimes(2);
+      expect(dispatchEvent).toHaveBeenCalledWith(
+        new Event('eip6963:requestProvider'),
+      );
+      expect(dispatchEvent).toHaveBeenCalledWith(
+        new CustomEvent('eip6963:announceProvider'),
+      );
```
jiexi: should we be explicit with call order?

jiexi:
```
-------------------------------------|---------|----------|---------|---------|---------------------------------------------------------
File                                 | % Stmts | % Branch | % Funcs | % Lines | Uncovered Line #s                                       
-------------------------------------|---------|----------|---------|---------|---------------------------------------------------------
All files                            |   65.97 |    63.94 |   64.89 |   65.88 |                                                         
 src                                 |   63.59 |    60.84 |   63.33 |   63.42 |                                                         
  BaseProvider.ts                    |   80.76 |    64.28 |     100 |   80.58 | 234,340-349,374-375,404-408,413-418,427,442-443,465-468 
  EIP6963.ts                         |   94.28 |    90.47 |     100 |   94.28 | 61,105                                                  
  MetaMaskInpageProvider.ts          |   63.54 |     73.8 |      50 |   64.21 | 138-140,146-158,187-188,202-211,232,260-272,400-433     
  StreamProvider.ts                  |   66.66 |    47.82 |   83.33 |   66.66 | 98,100,103-109,135,154-164,189-193                      
  constants.ts                       |     100 |      100 |     100 |     100 |                                                         
  initializeInpageProvider.ts        |       0 |        0 |       0 |       0 | 3-93                                                    
  messages.ts                        |   64.28 |      100 |   58.33 |   64.28 | 6-8,13,18-20                                            
  shimWeb3.ts                        |       0 |        0 |       0 |       0 | 11-61                                                   
  siteMetadata.ts                    |    9.09 |        0 |       0 |    9.67 | 16-115                                                  
  utils.ts                           |   96.77 |    85.71 |     100 |      96 | 51                                                      
 src/extension-provider              |   89.47 |       80 |     100 |   89.47 |                                                         
  createExternalExtensionProvider.ts |   89.47 |       80 |     100 |   89.47 | 43-44                                                   
 src/middleware                      |     100 |      100 |     100 |     100 |                                                         
  createRpcWarningMiddleware.ts      |     100 |      100 |     100 |     100 |                                                         
-------------------------------------|---------|----------|---------|---------|---------------------------------------------------------
Jest: "global" coverage threshold for statements (66.43%) not met: 65.97%
Jest: "global" coverage threshold for branches (64.25%) not met: 63.94%
Jest: "global" coverage threshold for lines (66.35%) not met: 65.88%
```

This only introduces two untested lines. Is it reasonable to adjust the global coverage to match?

jiexi:
Ticket: [mmp-869](https://github.com/MetaMask/MetaMask-planning/issues/869)
Related: https://github.com/MetaMask/providers/pull/263
Related: https://github.com/MetaMask/test-dapp/pull/243

# Screenshot
Screenshot 2023-07-06 at 1 55 58 PM

# Recording
* Coinbase Wallet "wins" the `window.ethereum` injection race
* MetaMask Extension is announced via EIP6963
https://github.com/MetaMask/test-dapp/assets/918701/6e62f97d-9908-4cc9-9e97-63b06534aada

jest.setup.browser.js
```diff
@@ -0,0 +1,87 @@
...
+
+    // Add to default key array to prevent removal during reset
+    sideEffects[obj].keys.push('addEventListener');
+
+    // Replace addEventListener with mock
+    global[obj].addEventListener = addEventListenerSpy;
+  });
+});
+
+// Reset JSDOM. This attempts to remove side effects from tests, however it does
```
Gudahtt, line 51: I ran into JSDOM state issues when working on the phishing warning page too. In that case, I found it easiest to just use a real browser. See https://github.com/MetaMask/phishing-warning/pull/62

Likewise for `browser-passworder`, that one is a plain library that we wanted to test in a real browser to ensure we were using browser APIs correctly. That uses playwright too: https://github.com/MetaMask/browser-passworder/blob/f870292ec5db729c8f69bbe4b2d3cff60053f333/test/index.spec.ts

Something to consider, if this needs further maintenance or continues to be a burden.

jiexi:
This should be caught up to spec now

src/initializeInpageProvider.ts
```diff
@@ -12,6 +13,11 @@ type InitializeProviderOptions = {
    */
   connectionStream: Duplex;
 
+  /**
+   * The EIP-6963 provider info that should be announced if set.
+   */
+  shouldAnnounceProviderInfo?: EIP6963ProviderInfo;
```
rekmarks: nit: IIRC this is contrary to our naming conventions since it's not a boolean, and we ought to change it to:

```suggestion
  /**
   * The EIP-6963 provider info that should be announced if set.
   */
  providerInfo?: EIP6963ProviderInfo;
```

jiexi: done: https://github.com/MetaMask/providers/pull/263/commits/c1d5dc6af5db4f8d1c8c2970815f5e7e339ead08

src/EIP6963.ts
jiexi: Thoughts here? @rekmarks @Mrtenz

src/EIP6963.ts
```diff
@@ -0,0 +1,199 @@
...
+      }
+      handleProvider(event.detail);
+    },
+  );
+
+  window.dispatchEvent(new Event(EIP6963EventNames.Request));
+}
+
+// https://caniuse.com/js-regexp-lookbehind
+// MetaMask Mobile, which uses Safari for the in-app browser, does NOT support lookbehind regexes.
```
jiexi: Jumping ahead and trying to get the unreleased EIP6963 provider changes into mobile-provider into mobile. Was getting this error: `SyntaxError: Invalid regular expression: invalid group specifier name` which seems to be a safari specific issue for regexes. This was the result of having a lookbehind match in the FQDN_REGEX which I have replaced and added this comment here, but maybe this should go somewhere else?

.eslintrc.js
```diff
@@ -26,12 +26,32 @@ module.exports = {
...
+      rules: {
+        // We're mixing Node and browser environments in these files.
+        'no-restricted-globals': 'off',
+      },
+    },
+
+    {
+      files: ['jest.setup.browser.js'],
+      env: { browser: true },
+      // This file contains copypasta and we're not going to bother fixing these.
```
mcmire, line 47: 😅

src/EIP6963.ts
```diff
@@ -0,0 +1,199 @@
...
+ * {@link EIP6963AnnounceProviderEvent}, and dispatches an
+ * {@link EIP6963RequestProviderEvent}.
+ *
+ * @param handleProvider - A function that handles an announced provider.
+ */
+export function requestProvider<HandlerReturnType>(
+  handleProvider: (providerDetail: EIP6963ProviderDetail) => HandlerReturnType,
+): void {
+  window.addEventListener(
+    EIP6963EventNames.Announce as any,
```
mcmire: What's the complaint from TypeScript here?

src/EIP6963.ts
mcmire: Do you want to take these two comments out, then, as you're no longer using lookbehinds?

src/EIP6963.ts
```diff
@@ -0,0 +1,199 @@
...
+/**
+ * Validates an {@link EIP6963RequestProviderEvent} object.
+ *
+ * @param event - The {@link EIP6963RequestProviderEvent} to validate.
+ * @returns Whether the {@link EIP6963RequestProviderEvent} is valid.
+ */
+function isValidRequestProviderEvent(
+  event: unknown,
+): event is EIP6963RequestProviderEvent {
+  const providerEvent = event as EIP6963RequestProviderEvent;
```
mcmire: Do you need this here? It seems that the `instanceof` check below is sufficient.

src/EIP6963.ts
```diff
@@ -0,0 +1,199 @@
...
+/**
+ * Validates an {@link EIP6963AnnounceProviderEvent} object.
+ *
+ * @param event - The {@link EIP6963AnnounceProviderEvent} to validate.
+ * @returns Whether the {@link EIP6963AnnounceProviderEvent} is valid.
+ */
+function isValidAnnounceProviderEvent(
+  event: unknown,
+): event is EIP6963AnnounceProviderEvent {
+  const providerEvent = event as EIP6963AnnounceProviderEvent;
```
mcmire: Same for this.

src/EIP6963.ts
```diff
@@ -0,0 +1,199 @@
...
+  providerDetail: unknown,
+): providerDetail is EIP6963ProviderDetail {
+  if (
+    !isObject(providerDetail) ||
+    !isObject(providerDetail.info) ||
+    !isObject(providerDetail.provider)
+  ) {
+    return false;
+  }
+  const { info } = providerDetail as EIP6963ProviderDetail;
```
mcmire: Same as above.

src/EIP6963.test.ts
```diff
@@ -0,0 +1,222 @@
...
+          const providerDetail = { info: getProviderInfo(), provider };
+
+          expect(() => announceProvider(providerDetail)).toThrow(
+            providerInfoValidationError(),
+          );
+        });
+      });
+    });
+
+    it('should announce a provider', () => {
```
mcmire: This tests dispatching the `announce` event, but should we also add a test for listening to the `request` event?

src/EIP6963.test.ts
```diff
@@ -0,0 +1,222 @@
...
+    });
+  });
+
+  describe('requestProvider', () => {
+    it('should receive an announced provider (called before announceProvider)', async () => {
+      const provider: any = { name: 'test' };
+      const providerDetail = { info: getProviderInfo(), provider };
+      const handleProvider = jest.fn();
+      const dispatchEvent = jest.spyOn(window, 'dispatchEvent');
+      const addEventListener = jest.spyOn(window, 'addEventListener');
```
mcmire: This test is a bit confusing to read, because I'm not sure what we're testing: are we testing what `requestProvider` does, or what `announceProvider` does? Or both?

Maybe this test ought to be in the group for `announceProvider`? Or, if you do want to test everything together, maybe it's not valuable to have tests grouped by function in this file, and just have a flatter list of tests?

src/EIP6963.ts
```diff
@@ -0,0 +1,199 @@
...
+
+/**
+ * Represents a provider and the information relevant for the dapp.
+ */
+export type EIP6963ProviderDetail = {
+  info: EIP6963ProviderInfo;
+  provider: BaseProvider;
+};
+
+// Requesting an EVM provider
```
mcmire: Should these be in JSDoc format?

src/index.ts
```diff
@@ -20,4 +35,6 @@ export {
   setGlobalProvider,
   shimWeb3,
   StreamProvider,
+  eip6963AnnounceProvider,
```
mcmire, line 38: Hmm, this name seems a bit mashed together. Too bad we can't put a type under a namespace of some kind.

src/EIP6963.ts
```diff
@@ -0,0 +1,199 @@
...
+};
+
+// Annoucing an EVM provider
+export type EIP6963AnnounceProviderEvent = CustomEvent & {
+  type: EIP6963EventNames.Announce;
+  detail: EIP6963ProviderDetail;
+};
+
+/**
+ * Forwards every announced provider to the provided handler by listening for
```
mcmire: Is it useful to indicate that this function is intended for dapps to use, as indicated in https://eips.ethereum.org/EIPS/eip-6963#window-events?

src/EIP6963.ts
```diff
@@ -0,0 +1,199 @@
...
+// https://github.com/thenativeweb/uuidv4/blob/bdcf3a3138bef4fb7c51f389a170666f9012c478/lib/uuidv4.ts#L5
+const UUID_V4_REGEX =
+  /(?:^[a-f0-9]{8}-[a-f0-9]{4}-4[a-f0-9]{3}-[a-f0-9]{4}-[a-f0-9]{12}$)|(?:^0{8}-0{4}-0{4}-0{4}-0{12}$)/u;
+
+// https://stackoverflow.com/a/20204811
+const FQDN_REGEX =
+  /(?=^.{4,253}$)(^((?!-)[a-zA-Z0-9-]{0,62}[a-zA-Z0-9]\.)+[a-zA-Z]{2,63}$)/u;
+
+/**
+ * Announces a provider by dispatching an {@link EIP6963AnnounceProviderEvent}, and
```
mcmire: Is it useful to indicate that this is for wallets to use, as indicated in https://eips.ethereum.org/EIPS/eip-6963#window-events?

mcmire (commented):
I read over EIP-6963, and the implementation makes sense. Just had minor questions about tests and type assertions.

src/EIP6963.ts
jiexi:
```
src/EIP6963.ts:61:3 - error TS2769: No overload matches this call.
  Overload 1 of 2, '(type: keyof WindowEventMap, listener: (this: Window, ev: Event | ErrorEvent | MessageEvent<any> | DeviceMotionEvent | DeviceOrientationEvent | ... 22 more ... | StorageEvent) => any, options?: boolean | ... 1 more ... | undefined): void', gave the following error.
    Argument of type 'EIP6963EventNames.Announce' is not assignable to parameter of type 'keyof WindowEventMap'.
  Overload 2 of 2, '(type: string, listener: EventListenerOrEventListenerObject, options?: boolean | AddEventListenerOptions | undefined): void', gave the following error.
    Argument of type '(event: EIP6963AnnounceProviderEvent) => void' is not assignable to parameter of type 'EventListenerOrEventListenerObject'.
      Type '(event: EIP6963AnnounceProviderEvent) => void' is not assignable to type 'EventListener'.
        Types of parameters 'event' and 'evt' are incompatible.
          Type 'Event' is not assignable to type 'EIP6963AnnounceProviderEvent'.
            Type 'Event' is missing the following properties from type 'CustomEvent<any>': detail, initCustomEvent
```

jiexi: fixed by via https://stackoverflow.com/a/47171216

src/EIP6963.ts
jiexi: done: https://github.com/MetaMask/providers/pull/263/commits/82abfbfa62b5f01b26b8945dad04af9eab5cb4b5

src/EIP6963.ts
jiexi: done: https://github.com/MetaMask/providers/pull/263/commits/82abfbfa62b5f01b26b8945dad04af9eab5cb4b5

src/EIP6963.ts
jiexi: done: https://github.com/MetaMask/providers/pull/263/commits/82abfbfa62b5f01b26b8945dad04af9eab5cb4b5

src/EIP6963.ts
jiexi: done: https://github.com/MetaMask/providers/pull/263/commits/82abfbfa62b5f01b26b8945dad04af9eab5cb4b5

src/EIP6963.ts
jiexi: done: https://github.com/MetaMask/providers/pull/263/commits/82abfbfa62b5f01b26b8945dad04af9eab5cb4b5

src/EIP6963.test.ts
jiexi: let's test it all together.

There's two scenarios:
* Provider is ready before Dapp
* Dapp is ready before Provider

Will try to organize the test around that

src/EIP6963.ts
```diff
@@ -0,0 +1,199 @@
...
+  );
+}
+
+/**
+ * Throws an error with link to EIP-6963 specifications.
+ *
+ * @param message - The message to include.
+ * @throws a friendly error with a link to EIP-6963.
+ */
+function throwErrorEIP6963(message: string) {
```
mcmire, line 206: For some reason, TypeScript doesn't know that this function doesn't actually do anything, so it assigns a return type of `void`. It would be more appropriate to assign `never` here so that TypeScript knows that once this is called, an invalid state is produced. Primarily, this will help the type guard in `requestProvider` do what it's supposed to do. **EDIT:** I guess this is not so important given your changes, but if we ever have problems with this again then that's the fix.

src/EIP6963.ts
mcmire: Okay. This makes sense, actually. We shouldn't expect that the type of object that `addEventListener` is given is automatically what we want, even if the event is custom.

~I've experimented with this a bit. It looks like you can change the function signature to `event: Event`, however, even after the type guard `isValidAnnouncementProviderEvent` is called, TypeScript still thinks `event` is an Event. I think this is happening because TypeScript doesn't know that `throwErrorEIP6963` throws an error. If you not only use `event: Event` but mark `throwErrorEIP6963` as `never` then you shouldn't need this type assertion.~ Oh I saw your changes, cool.

src/EIP6963.test.ts
jiexi: done: https://github.com/MetaMask/providers/pull/263/commits/06ce8e5ef4c35c376a72ae0bb1256bf12c37c265

src/EIP6963.test.ts
jiexi: flattened tests here: https://github.com/MetaMask/providers/pull/263/commits/06ce8e5ef4c35c376a72ae0bb1256bf12c37c265

src/EIP6963.ts
jiexi: done: https://github.com/MetaMask/providers/pull/263/commits/c38d4e4d3c96d16e2fb5103ea11cddb70aa904ed

src/EIP6963.ts
jiexi: done: https://github.com/MetaMask/providers/pull/263/commits/8641b8fff70f7c6c090837c8d7a5fc4df5bad631

src/EIP6963.ts
jiexi: done: https://github.com/MetaMask/providers/pull/263/commits/8641b8fff70f7c6c090837c8d7a5fc4df5bad631

src/index.ts
jiexi: hmm.. i'm not sure it would make sense to drop the EIP6963 prefixes either 😬

.eslintrc.js
jiexi:
```
/Users/jiexi/Projects/providers/jest.setup.browser.js
  25:1  error  All hooks must be wrapped in a describe block  jest/require-top-level-describe
  30:5  error  Missing JSDoc block description                jsdoc/require-description
  32:1  error  Missing JSDoc @param "type" description        jsdoc/require-param-description
  32:1  error  Missing JSDoc @param "type" type               jsdoc/require-param-type
  33:1  error  Missing JSDoc @param "listener" description    jsdoc/require-param-description
  33:1  error  Missing JSDoc @param "listener" type           jsdoc/require-param-type
  34:1  error  Missing JSDoc @param "options" description     jsdoc/require-param-description
  34:1  error  Missing JSDoc @param "options" type            jsdoc/require-param-type
  55:1  error  All hooks must be wrapped in a describe block  jest/require-top-level-describe
```

jiexi: Definitely need to ignore `jest/require-top-level-describe`

jiexi: do you feel it's worthwhile to add jsdoc to the rest? I can go either way

rekmarks: lol, this was me btw. It seems strictly not worth it, as I have a hard time imagining a lot of work being done in that file, but not my call of course!

mcmire: I'm good with this, just thought it was funny :)

mcmire (approved):
Looks good to me!

PR #264, opened 2023-06-22
warn wallet_watchAsset ERC721,ERC1155

jiexi:
Adds a console.warn when `wallet_watchAsset` is called with `params: {type: 'ERC721'}` or `params: {type: 'ERC1155'}`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,7 +37,6 @@
     "test:watch": "jest --watch"
   },
   "dependencies": {
-    "@metamask/controller-utils": "^4.0.0",
     "@metamask/object-multiplex": "^1.1.0",
     "@metamask/safe-event-emitter": "^3.0.0",
     "detect-browser": "^5.2.0",
@@ -94,10 +93,6 @@
     "registry": "https://registry.npmjs.org/"
   },
   "lavamoat": {
-    "allowScripts": {
-      "@metamask/controller-utils>babel-runtime>core-js": false,
-      "@metamask/controller-utils>ethereumjs-util>ethereum-cryptography>keccak": false,
-      "@metamask/controller-utils>ethereumjs-util>ethereum-cryptography>secp256k1": false
-    }
+    "allowScripts": {}
   }
 }
--- a/package.json
+++ b/package.json
@@ -37,6 +37,7 @@
     "test:watch": "jest --watch"
   },
   "dependencies": {
+    "@metamask/controller-utils": "^4.0.0",
     "@metamask/object-multiplex": "^1.1.0",
     "@metamask/safe-event-emitter": "^3.0.0",
     "detect-browser": "^5.2.0",
--- a/package.json
+++ b/package.json
@@ -94,6 +94,10 @@
     "registry": "https://registry.npmjs.org/"
   },
   "lavamoat": {
-    "allowScripts": {}
+    "allowScripts": {
+      "@metamask/controller-utils>babel-runtime>core-js": false,
+      "@metamask/controller-utils>ethereumjs-util>ethereum-cryptography>keccak": false,
+      "@metamask/controller-utils>ethereumjs-util>ethereum-cryptography>secp256k1": false
+    }
   }
 }
--- a/src/messages.ts
+++ b/src/messages.ts
@@ -37,6 +37,7 @@ const messages = {
     rpc: {
       ethDecryptDeprecation: `MetaMask: The RPC method 'eth_decrypt' is deprecated and may be removed in the future.\nFor more information, see: https://medium.com/metamask/metamask-api-method-deprecation-2b0564a84686`,
       ethGetEncryptionPublicKeyDeprecation: `MetaMask: The RPC method 'eth_getEncryptionPublicKey' is deprecated and may be removed in the future.\nFor more information, see: https://medium.com/metamask/metamask-api-method-deprecation-2b0564a84686`,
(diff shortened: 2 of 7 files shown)


src/middleware/createRpcWarningMiddleware.ts
```diff
@@ -27,6 +28,15 @@ export function createRpcWarningMiddleware(
     ) {
       log.warn(messages.warnings.rpc.ethGetEncryptionPublicKeyDeprecation);
       sentWarnings.ethGetEncryptionPublicKeyDeprecation = true;
+    } else if (
+      !sentWarnings.walletWatchAssetNFTExperimental &&
+      req.method === 'wallet_watchAsset' &&
+      ['ERC721', 'ERC1155'].includes(
```
jiexi: should these get pulled in as constants from controller-utils?

src/messages.ts
```diff
@@ -37,6 +37,7 @@ const messages = {
     rpc: {
       ethDecryptDeprecation: `MetaMask: The RPC method 'eth_decrypt' is deprecated and may be removed in the future.\nFor more information, see: https://medium.com/metamask/metamask-api-method-deprecation-2b0564a84686`,
       ethGetEncryptionPublicKeyDeprecation: `MetaMask: The RPC method 'eth_getEncryptionPublicKey' is deprecated and may be removed in the future.\nFor more information, see: https://medium.com/metamask/metamask-api-method-deprecation-2b0564a84686`,
+      walletWatchAssetNFTExperimental: `MetaMask: The RPC method 'wallet_watchAsset' is experimental for ERC721/ERC1155 assets and may change in the future.\nFor more information, see: https://TODO`,
```
jiexi: what should this messaging be?

shanejonas: `see: https://docs.metamask.io/wallet/reference/wallet_watchasset/` maybe?

adonesky1: maybe we link to [MIP-1](https://github.com/MetaMask/metamask-improvement-proposals/blob/main/MIPs/mip-1.md) and the [proposal lifecycle](https://github.com/MetaMask/metamask-improvement-proposals/blob/main/PROCESS-GUIDE.md#proposal-lifecycle) in the process-guide document? cc @vandan

src/middleware/createRpcWarningMiddleware.ts
```diff
@@ -27,6 +28,15 @@ export function createRpcWarningMiddleware(
...
+        (req as JsonRpcRequest<{ type: string }>).params?.type || '',
```
adonesky1, line 36: do we need `|| ''` here?

adonesky1 (commented):
One small question, and we just need to update that warning message and then I think we're good to go!

src/middleware/createRpcWarningMiddleware.ts
jiexi: `Array<string>.includes` expects only `string` type as an arg. it's potentially undefined due to the `?` optional chaining

adonesky1: ah ok makes sense

src/messages.ts
vandan: Yes, I think both of those links would be good references.

jiexi:
[@metamask/controller-utils@4.0.0](https://socket.dev/npm/package/@metamask/controller-utils/overview/4.0.0) is used in extension. Okay if we `@SocketSecurity ignore-all` the socket-security alerts?

src/middleware/createRpcWarningMiddleware.test.ts
```diff
@@ -79,51 +112,56 @@ describe('createRpcWarningMiddleware', () => {
...
-    it('should not issue a warning', async () => {
-      const consoleWarnSpy = jest.spyOn(globalThis.console, 'warn');
-      const middleware = createRpcWarningMiddleware(globalThis.console);
-      const engine = new JsonRpcEngine();
-      engine.push(middleware);
-
-      await engine.handle({ jsonrpc: '2.0', id: 1, method: 'eth_chainId' });
+  for (const { method, params = {}, scenario } of unaffected) {
+    const name = scenario ? `${method} (${scenario})` : method;
+    describe(`${name}`, () => {
```
adonesky1: [nit] looks like you could use [describe.each](https://jestjs.io/docs/api#1-describeeachtablename-fn-timeout) here.

package.json
```diff
@@ -37,6 +37,7 @@
     "test:watch": "jest --watch"
   },
   "dependencies": {
+    "@metamask/controller-utils": "^4.0.0",
```
adonesky1: All of the added sub-deps here will be deduped in the extension but I'm wondering its worth making this package so much heavier just for those asset type strings... Thoughts @mcmire @Gudahtt ?

adonesky1: I realize I originally said yes when you asked if we should pull the consts in this way...

jiexi: same thoughts about if it is worth including the package

pedronfigueiredo (dismissed):

package.json
mcmire: Hmm, I also have the same qualms. We should definitely try to make sure that our libraries are as small as possible on their own. Perhaps we just copy over those constants for now.

adonesky1 (approved):
LGTM!

PR #275, opened 2023-08-24
Use `@metamask/json-rpc-engine`

legobeat:
This validates if https://github.com/MetaMask/rpc-errors/pull/102 fixes type error in https://github.com/MetaMask/providers/actions/runs/5554034666/job/15044724731.

PR #253, opened 2023-04-19
Use `@metamask/json-rpc-engine` and `@metamask/rpc-errors`

FrederikBolding:

PR #276, opened 2023-08-24
test using @metamask/utils@8.0.0

legobeat:
This is to validate if https://github.com/MetaMask/utils/pull/130 fixes the [type error](https://github.com/MetaMask/providers/actions/runs/5959652471/job/16165741468?pr=275) in https://github.com/MetaMask/providers/pull/275, which this is based on.

Original PR: #253


legobeat:
#277

PR #280, opened 2023-09-08
Make `chainId`, `networkVersion`, `selectedAddress` readonly and log deprecation warning

jiexi:
Currently, it's possible to access the properties `chainId`, `networkVersion`, and `selectedAddress` on the provider object. These are not part of the [EIP-1193](https://eips.ethereum.org/EIPS/eip-1193) spec. More specifically, we are trying to remove networkId where possible and `networkVersion` currently poses an issue as it references networkId and that its removal may break some dapps if they are accessing this non-documented property. The plan is to add a deprecation warning when accessing any of these properties, and to fully remove public access to them at a later point.

We achieve this by making the actual properties private and putting them behind getter functions that ensure a console warning is issued on access. Additionally, we make the properties readonly by not providing a setter.

* See: https://github.com/MetaMask/MetaMask-planning/issues/1068

## Examples

Screenshot 2023-09-08 at 12 15 01 PM

### DIFF
--- a/src/initializeInpageProvider.ts
+++ b/src/initializeInpageProvider.ts
@@ -55,6 +55,11 @@ export function initializeProvider({
   const proxiedProvider = new Proxy(provider, {
     // some common libraries, e.g. web3@1.x, mess with our API
     deleteProperty: () => true,
+    // fix issue with Proxy unable to access private variables from getters
+    // https://stackoverflow.com/a/73051482
+    get(target, propName: 'chainId' | 'networkVersion' | 'selectedAddress') {
+      return target[propName];
+    },
   });
 
   if (shouldSetOnWindow) {
--- a/src/messages.ts
+++ b/src/messages.ts
@@ -24,6 +24,10 @@ const messages = {
       `MetaMask: Connected to chain with ID "${chainId}".`,
   },
   warnings: {
+    // deprecated properties
+    chainIdDeprecation: `MetaMask: 'ethereum.chainId' is deprecated and may be removed in the future. Please use the 'eth_chainId' RPC method instead.\nFor more information, see: TODO`,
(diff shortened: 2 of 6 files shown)


src/messages.ts
```diff
@@ -24,6 +24,10 @@ const messages = {
       `MetaMask: Connected to chain with ID "${chainId}".`,
   },
   warnings: {
+    // deprecated properties
+    chainIdDeprecation: `MetaMask: 'ethereum.chainId' is deprecated and may be removed in the future. Please use the 'eth_chainId' RPC method instead.\nFor more information, see: TODO`,
+    networkVersionDeprecation: `MetaMask: 'ethereum.networkVersion' is deprecated and may be removed in the future. Please use the 'net_version' RPC method instead.\nFor more information, see: TODO`,
+    selectedAddressDeprecation: `MetaMask: 'ethereum.selectedAddress' is deprecated and may be removed in the future. Please use the 'eth_accounts' RPC method instead.\nFor more information, see: TODO`,
```
jiexi: @vandan need links here

jiexi:
@metamaskbot publish-preview

src/initializeInpageProvider.ts
```diff
@@ -55,6 +55,11 @@ export function initializeProvider({
   const proxiedProvider = new Proxy(provider, {
     // some common libraries, e.g. web3@1.x, mess with our API
     deleteProperty: () => true,
+    // fix issue with Proxy unable to access private variables from getters
+    // https://stackoverflow.com/a/73051482
+    get(target, propName: 'chainId' | 'networkVersion' | 'selectedAddress') {
+      return target[propName];
+    },
```
jiexi, line 62: feels hacky. would love thoughts here

src/messages.ts
jiexi: https://github.com/MetaMask/metamask-improvement-proposals/discussions/23 maybe

legobeat (approved):
I don't see why not and implementation LGTM

PR #306, opened 2024-03-06
Remove deprecated properties, networkChanged event, and offline send() net_version support

jiexi:
This PR removes properties that are not officially documented and should not have been exposed in the first place. It also removes support for the networkChange event and offline `net_version` resolution which will allow us to remove the last remaining chunk of networkId usage in Extension. These have all been properly deprecated, the most recent deprecations being 180 days old now.

Removes the following:
- window.ethereum.chainId
- window.ethereum.networkVersion
- window.ethereum.selectedAddress
- networkChanged event
- offline net_version support in window.ethereum.send()

See: https://github.com/MetaMask/MetaMask-planning/issues/2195
See: https://github.com/MetaMask/metamask-improvement-proposals/discussions/23
See: https://github.com/MetaMask/MetaMask-planning/issues/2176
See: https://github.com/MetaMask/metamask-extension/pull/23375

### **Before**

Screenshot 2024-03-07 at 10 16 08 AM

### **After**

Screenshot 2024-03-07 at 10 23 54 AM


src/MetaMaskInpageProvider.ts
```diff
@@ -167,31 +150,15 @@ export class MetaMaskInpageProvider extends AbstractStreamProvider {
...
+    throw new Error(messages.errors.invalidPropertyChainId());
   }
 
   get selectedAddress(): string | null {
-    if (!this._sentWarnings.selectedAddress) {
-      this._log.warn(messages.warnings.selectedAddressDeprecation);
-      this._sentWarnings.selectedAddress = true;
-    }
-    return super.selectedAddress;
+    throw new Error(messages.errors.invalidPropertySelectedAddress());
```
jiexi: These getter overrides must exist in order to prevent publicly exposing them via inheritance from the BaseProvider. We cannot just make those properties private in the BaseProvider themselves because the subclasses still need access to them.

src/StreamProvider.test.ts
```diff
@@ -397,85 +392,10 @@ describe('StreamProvider', () => {
           mockStream.notify(mockStreamName, {
             jsonrpc: '2.0',
             method: 'metamask_chainChanged',
-            params: { chainId: '0x1', networkVersion: '0x1' },
+            params: { chainId: '0x1' },
           });
         });
       });
-
-      it('handles chain changes with intermittent disconnection', async () => {
```
jiexi, line 405: we don't have networkId anymore, nor the notion of temporary disconnection/switching when networkId is set to `loading`

src/StreamProvider.ts
```diff
@@ -163,43 +159,30 @@ export abstract class AbstractStreamProvider extends BaseProvider {
...
-        networkVersion,
       });
       return;
     }
 
-    if (networkVersion === 'loading') {
-      this._handleDisconnect(true);
-    } else {
-      super._handleChainChanged({ chainId });
-    }
```
jiexi, line 202: we don't have networkId anymore, nor the notion of temporary disconnection/switching when networkId is set to `loading`

shanejonas:
Total sidebar here but



Screenshot 2024-03-06 at 3 43 05 PM

> null {id: undefined, jsonrpe: '2.0', result: '100'}

Wow looks like `send` doesnt generate an ID like how we have `request` set up. I do like that `send` leaves it up to the client to deal with the ID.

shanejonas (dismissed):

src/MetaMaskInpageProvider.ts
legobeat: Could they be made `protected`, if anything?

src/MetaMaskInpageProvider.ts
```diff
@@ -167,31 +150,15 @@ export class MetaMaskInpageProvider extends AbstractStreamProvider {
...
 
   get chainId(): string | null {
-    if (!this._sentWarnings.chainId) {
-      this._log.warn(messages.warnings.chainIdDeprecation);
-      this._sentWarnings.chainId = true;
-    }
-    return super.chainId;
-  }
-
-  get networkVersion(): string | null {
```
adonesky1, line 181: any reason we're not throwing a `new Error(messages.errors.invalidPropertyNetworkVersion())` for this one?

jiexi: Mostly because networkVersion doesn't require a getter override, so it's possible to remove it completely without issue unlike the chainId and selectedAddress getters. I am also not throwing when `networkChanged` are listened to.

src/MetaMaskInpageProvider.ts
jiexi: I tried making the BaseProvider getters `protected` and removing the overrides in MetaMaskInpageProvider, but the properties are still accessible. If i add `protected` to the MetaMaskInpageProvider getters, it does nothing and the properties are also still accessible

src/initializeInpageProvider.ts
```diff
@@ -62,11 +62,6 @@ export function initializeProvider({
   const proxiedProvider = new Proxy(provider, {
```
BelfordZ, line 62: its not clear to me why this is a proxy, nor what the deleteProperty problem is.

When we use this in extension, the return value is never used, and I dont see anywhere that we are setting a new target, but maybe Im missing something. [Heres the extension place im referring to.](https://github.com/MetaMask/metamask-extension/blob/develop/app/scripts/inpage.js#L59)

BelfordZ (dismissed):

src/initializeInpageProvider.ts
jiexi: I believe it's a proxy to override `deleteProperty: () => true,`

src/MetaMaskInpageProvider.ts
mcmire: Hmm, I agree. If we only throw for some of the things we've removed but not others, then couldn't it possibly be surprising?

mcmire (dismissed):
One comment, but this change makes sense to me regardless.

src/MetaMaskInpageProvider.ts
jiexi: Throwing error for networkVersion here https://github.com/MetaMask/providers/pull/306/commits/10ee3a398fb87a479e656474e0fd80339e4eb96c

adonesky1 (approved):

PR #319, opened 2024-05-16
fix(deps): move webextension-polyfill to peerDeps + devDeps

legobeat:
The `webextension-polyfill` package is only used once for a type-import and assumed in environment. This makes it a `peerDependency` and a `devDependency` instead of a `dependency`. A benefit of this is that users can start using new version without us having to release explicit support for every version bump, besides from being more correct.

`peerDependencies` version is aligning with already used transitive via `extension-port-stream`.

This also upgrades the used version from `webextension-polyfill@0.10.0` to `webextension-polyfill@0.12.0`.

---

### Related
- https://github.com/MetaMask/extension-port-stream/pull/54

### DIFF
--- a/package.json
+++ b/package.json
@@ -67,8 +67,7 @@
     "extension-port-stream": "^3.0.0",
     "fast-deep-equal": "^3.1.3",
     "is-stream": "^2.0.0",
-    "readable-stream": "^3.6.2",
-    "webextension-polyfill": "^0.10.0"
+    "readable-stream": "^3.6.2"
   },
   "devDependencies": {
     "@lavamoat/allow-scripts": "^3.0.0",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -105,7 +104,11 @@
     "ts-node": "^10.7.0",
     "tsup": "^7.2.0",
     "typedoc": "^0.23.15",
-    "typescript": "~4.8.4"
+    "typescript": "~4.8.4",
+    "webextension-polyfill": "^0.12.0"
+  },
+  "peerDependencies": {
+    "webextension-polyfill": ">=0.10.0 <1.0"
```
rekmarks: Although this aligns with `extension-port-stream`, I don't think it's correct. SemVer permits minor versions to be breaking for major version `0` ([ref](https://devhints.io/semver)). We should probably make this `"^0.12.0"` as well.

legobeat: While true, we at least know it works with past versions.

So if we want more of the tradeoff of "allow chilling on maintenance and not staying on top of every future update" (considering history here was part of the decision - we'd want days not weeks/months) vs "make sure future updates don't cause breakage" , then how about `^0.10 || ^0.11 || ^0.12`?

legobeat: One nuance: Using identical ranges will ensure identical resulting versions due to resolutions being identical. I don't think it would come up in practice here but defaulting to consistency does prevent some classes of unexpected behavior.

legobeat: I updated it to `^0.10.0 || ^0.11.0 || ^0.12.0` now.

Mrtenz (approved):

PR #323, opened 2024-05-29
feat: Modify provider RPC error log severity from `error` to `warn`

rekmarks:
The error middleware injected in the provider-side `json-rpc-engine` was calling `log.error()` (in practice, `log` = `console`) every time it observed a JSON-RPC error response.

We received reports that this error log is caught by the monitoring infra of some dapps. It is ultimately the caller's responsibility to handle errors where they call `ethereum.request()`, and they can always `console.error()` at that point themselves. In light of the above, we decrease the severity of our own log statement to `warn`.

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -61,7 +61,7 @@ function createErrorMiddleware(
       if (!error) {
         return done();
       }
-      log.debug(`MetaMask - RPC Error: ${error.message}`, error);
+      log.warn(`MetaMask - RPC Error: ${error.message}`, error);
       return done();
     });
   };
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -61,7 +61,7 @@ function createErrorMiddleware(
       if (!error) {
         return done();
       }
-      log.error(`MetaMask - RPC Error: ${error.message}`, error);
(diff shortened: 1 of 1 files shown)


src/utils.ts
```diff
@@ -61,7 +61,7 @@ function createErrorMiddleware(
       if (!error) {
         return done();
       }
-      log.error(`MetaMask - RPC Error: ${error.message}`, error);
+      log.debug(`MetaMask - RPC Error: ${error.message}`, error);
```
legobeat: Could `log.warn` do the trick? Would keep it on `stderr` for non-browser environment. Seems like it would also solve the reported situation by allowing consumers of errors to filter in the same way.

```suggestion
      log.warn(`MetaMask - RPC Error: ${error.message}`, error);
```

rekmarks: I think this is the most conservative change that will solve the reported issue. Alright, let's go with it 👍

rekmarks:
The stretch-goal version of this would be to make logging configurable via some method on the provider object (constructor arguments don't help most of our consumers), but that's definitely a stretch goal.

legobeat (approved):

legobeat:
@rekmarks Going ahead and merging this to get it into next release FYI

PR #340, opened 2024-06-29
fix: ts-bridge import json error

### LINKED ISSUE #339: bundler file .ext error
The json file should not endswith `.mjs`
image

image

image

Innei:
ts-bridge converts json imports to .json.cjs or .json.mjs after compilation, which is incorrect behavior.

Fixes https://github.com/MetaMask/providers/issues/339


Mrtenz:
Good catch, that's definitely not intended in `ts-bridge`. I will fix this upstream next week.

MajorLift:
Closed by https://github.com/MetaMask/providers/pull/343

PR #390, opened 2024-11-27
fix: Allow optional `jsonRpcStreamName` for inpage provider

Gudahtt:
The `MetaMaskInpageProvider` type was mistakenly typed to require the `jsonRpcStreamName` as a constructor parameter, despite the author clearly intending this to be optional. This type error was fixed.

This type error motivated a different bug, which is that validation for `jsonRpcStreamName` was added to `initializeInapgeProvider` in #381, constituting an undocumented breaking change (currently blocking us from updating this package in `metamask-extension`). This validation has been removed now that it's no longer needed.

Fixes #389

PR #391, opened 2024-11-27
feat: Add new export for `initializeInpageProvider`

Gudahtt:
Add a new top-level export map entry for `initializeInpageProvider ` that maps correctly onto CJS and ESM builds, and add a fallback JavaScript file at the root of the repository for build systems that don't support export maps.

This gives the best of both worlds: build systems that support exports will behave correctly, and those that don't will get the CJS fallback.

The `metamask-extension` project uses two build systems: one which obeys export maps, one which ignores them (Webpack and browserify respectively). The existing `initializeInpageProvider` export map entry doesn't work for browserify because it's looking for `initializeInpageProvider.js`. The file extension is wrong. This fixes that problem.

This strategy was copied from https://github.com/MetaMask/snaps/blob/cbe55deb75e640c03add1ebf9235d4abd513e9d5/packages/snaps-sdk/jsx.js

### DIFF
--- a/package.json
+++ b/package.json
@@ -49,6 +49,10 @@
         "default": "./dist/initializeInpageProvider.cjs"
       }
     },
+    "./dist/initializeInpageProvider.cjs": {
+      "types": "./dist/initializeInpageProvider.d.cts",
+      "default": "./dist/initializeInpageProvider.cjs"
+    },
     "./stream-provider": {
       "import": {
         "types": "./dist/StreamProvider.d.mts",
--- a/package.json
+++ b/package.json
@@ -49,9 +49,15 @@
         "default": "./dist/initializeInpageProvider.cjs"
       }
     },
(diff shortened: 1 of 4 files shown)


Gudahtt:
Leaving this in draft until we get the next patch release out with this bug fix: https://github.com/MetaMask/providers/pull/390

package.json
```diff
@@ -49,6 +49,16 @@
         "default": "./dist/initializeInpageProvider.cjs"
       }
     },
+    "./initializeInpageProvider": {
```
Gudahtt, line 52: We can remove the other export entries that are no longer needed in a later PR. That would be a breaking change.

Tracked here: https://github.com/MetaMask/providers/issues/393

Mrtenz (approved):

Magkoooh (approved):

Magkoooh (approved):

PR #395, opened 2024-12-10
Implement Wallet Discovery For Multichain API #2970

ffmcgee725:
Addressing [Implement Wallet Discovery For Multichain API #2970](https://github.com/MetaMask/MetaMask-planning/issues/2970)

feat: support CAIP294 (standardized messaging transport for browser extension wallets)

feat: make sure we pass in extensionId when announcing wallet

### DIFF
--- a/src/CAIP294.test.ts
+++ b/src/CAIP294.test.ts
@@ -1,0 +1,189 @@
+import {
+  announceWallet,
+  CAIP294EventNames,
+  type CAIP294WalletData,
+  requestWallet,
+} from './CAIP294';
+
+const getWalletData = (): CAIP294WalletData => ({
+  uuid: '350670db-19fa-4704-a166-e52e178b59d2',
+  name: 'Example Wallet',
+  icon: 'data:image/svg+xml,<svg xmlns="http://www.w3.org/2000/svg"/>',
+  rdns: 'com.example.wallet',
+  extensionId: 'abcdefghijklmnopqrstuvwxyz',
+});
+
+const walletDataValidationError = () =>
+  new Error(
+    `Invalid CAIP-294 WalletData object received from ${CAIP294EventNames.Prompt}. See https://github.com/ChainAgnostic/CAIPs/blob/bc4942857a8e04593ed92f7dc66653577a1c4435/CAIPs/caip-294.md for requirements.`,
+  );
+
+describe('CAIP-294', () => {
+  describe('wallet data validation', () => {
+    it('throws if the wallet data is not a plain object', () => {
+      [null, undefined, Symbol('bar'), []].forEach((invalidInfo) => {
+        expect(() => announceWallet(invalidInfo as any)).toThrow(
+          walletDataValidationError(),
+        );
+      });
+    });
+
+    it('throws if the `icon` field is invalid', () => {
+      [
+        null,
+        undefined,
+        '',
+        'not-a-data-uri',
+        'https://example.com/logo.png',
+        'data:text/plain;blah',
+        Symbol('bar'),
+      ].forEach((invalidIcon) => {
+        const walletInfo = getWalletData();
+        walletInfo.icon = invalidIcon as any;
+
+        expect(() => announceWallet(walletInfo)).toThrow(
+          walletDataValidationError(),
+        );
+      });
+    });
+
+    it('throws if the `name` field is invalid', () => {
+      [null, undefined, '', {}, [], Symbol('bar')].forEach((invalidName) => {
+        const walletInfo = getWalletData();
+        walletInfo.name = invalidName as any;
+
+        expect(() => announceWallet(walletInfo)).toThrow(
+          walletDataValidationError(),
+        );
+      });
+    });
+
+    it('throws if the `uuid` field is invalid', () => {
+      [null, undefined, '', 'foo', Symbol('bar')].forEach((invalidUuid) => {
+        const walletInfo = getWalletData();
+        walletInfo.uuid = invalidUuid as any;
+
+        expect(() => announceWallet(walletInfo)).toThrow(
+          walletDataValidationError(),
+        );
+      });
+    });
+
+    it('throws if the `rdns` field is invalid', () => {
+      [
+        null,
+        undefined,
+        '',
+        'not-a-valid-domain',
+        '..com',
+        'com.',
+        Symbol('bar'),
+      ].forEach((invalidRdns) => {
+        const walletInfo = getWalletData();
+        walletInfo.rdns = invalidRdns as any;
+
+        expect(() => announceWallet(walletInfo)).toThrow(
+          walletDataValidationError(),
+        );
+      });
+    });
+
+    it('allows `extensionId` to be undefined or a string', () => {
+      const walletInfo = getWalletData();
+      expect(() => announceWallet(walletInfo)).not.toThrow();
+
+      delete walletInfo.extensionId;
+
+      expect(() => announceWallet(walletInfo)).not.toThrow();
+
+      walletInfo.extensionId = 'valid-string';
+      expect(() => announceWallet(walletInfo)).not.toThrow();
+    });
+  });
+
+  it('wallet is announced before dapp requests', async () => {
+    const walletData = getWalletData();
+    const handleWallet = jest.fn();
+    const dispatchEvent = jest.spyOn(window, 'dispatchEvent');
+    const addEventListener = jest.spyOn(window, 'addEventListener');
+
+    announceWallet(walletData);
+    requestWallet(handleWallet);
+    await delay();
+
+    expect(dispatchEvent).toHaveBeenCalledTimes(3);
+    expect(dispatchEvent).toHaveBeenNthCalledWith(
+      1,
+      new CustomEvent(CAIP294EventNames.Announce, expect.any(Object)),
+    );
+    expect(dispatchEvent).toHaveBeenNthCalledWith(
+      2,
+      new CustomEvent(CAIP294EventNames.Prompt, expect.any(Object)),
+    );
+    expect(dispatchEvent).toHaveBeenNthCalledWith(
+      3,
+      new CustomEvent(CAIP294EventNames.Announce, expect.any(Object)),
+    );
+
+    expect(addEventListener).toHaveBeenCalledTimes(2);
+    expect(addEventListener).toHaveBeenCalledWith(
+      CAIP294EventNames.Announce,
+      expect.any(Function),
+    );
+    expect(addEventListener).toHaveBeenCalledWith(
+      CAIP294EventNames.Prompt,
+      expect.any(Function),
+    );
+
+    expect(handleWallet).toHaveBeenCalledTimes(1);
+    expect(handleWallet).toHaveBeenCalledWith(
+      expect.objectContaining({ params: walletData }),
+    );
+  });
+
+  it('dapp requests before wallet is announced', async () => {
+    const walletData = getWalletData();
+    const handleWallet = jest.fn();
+    const dispatchEvent = jest.spyOn(window, 'dispatchEvent');
+    const addEventListener = jest.spyOn(window, 'addEventListener');
+
+    requestWallet(handleWallet);
+    announceWallet(walletData);
+    await delay();
+
+    expect(dispatchEvent).toHaveBeenCalledTimes(2);
+    expect(dispatchEvent).toHaveBeenNthCalledWith(
+      1,
+      new CustomEvent(CAIP294EventNames.Prompt, expect.any(Object)),
+    );
+    expect(dispatchEvent).toHaveBeenNthCalledWith(
+      2,
+      new CustomEvent(CAIP294EventNames.Announce, expect.any(Object)),
+    );
+
+    expect(addEventListener).toHaveBeenCalledTimes(2);
+    expect(addEventListener).toHaveBeenCalledWith(
+      CAIP294EventNames.Announce,
+      expect.any(Function),
+    );
+    expect(addEventListener).toHaveBeenCalledWith(
+      CAIP294EventNames.Prompt,
+      expect.any(Function),
+    );
+
+    expect(handleWallet).toHaveBeenCalledTimes(1);
+    expect(handleWallet).toHaveBeenCalledWith(
+      expect.objectContaining({ params: walletData }),
+    );
+  });
+});
+
+/**
+ * Delay for a number of milliseconds by awaiting a promise
+ * resolved after the specified number of milliseconds.
+ *
+ * @param ms - The number of milliseconds to delay for.
+ */
+async function delay(ms = 1) {
+  return new Promise((resolve) => setTimeout(resolve, ms));
+}
--- a/src/CAIP294.test.ts
+++ b/src/CAIP294.test.ts
@@ -101,6 +101,17 @@ describe('CAIP-294', () => {
     });
   });
 
+  it('throws if the `extensionId` field is invalid', () => {
+    [null, '', 42, Symbol('bar')].forEach((invalidExtensionId) => {
+      const walletInfo = getWalletData();
+      walletInfo.extensionId = invalidExtensionId as any;
+
+      expect(() => announceWallet(walletInfo)).toThrow(
+        walletDataValidationError(),
+      );
+    });
+  });
+
   it('wallet is announced before dapp requests', async () => {
     const walletData = getWalletData();
     const handleWallet = jest.fn();
--- a/src/CAIP294.ts
+++ b/src/CAIP294.ts
@@ -1,5 +1,6 @@
 import { isObject } from '@metamask/utils';
 
+import type { BaseProviderInfo } from './types';
 import { FQDN_REGEX, UUID_V4_REGEX } from './utils';
 
 /**
@@ -27,12 +28,8 @@ declare global {
  * @property rdns - The reverse syntax domain name identifier for the wallet.
  * @property extensionId - The canonical extension ID of the wallet provider for the active browser.
  */
-export type CAIP294WalletData = {
-  uuid: string;
-  name: string;
-  icon: string;
-  rdns: string;
-  extensionId?: string;
+export type CAIP294WalletData = BaseProviderInfo & {
+  extensionId?: string | undefined;
 };
 
 /**
--- a/src/CAIP294.ts
+++ b/src/CAIP294.ts
@@ -79,8 +79,27 @@ function isValidRequestWalletEvent(
     event instanceof CustomEvent &&
     event.type === CAIP294EventNames.Prompt &&
     isObject(event.detail) &&
-    event.detail.method === 'wallet_prompt'
+    event.detail.method === 'wallet_prompt' &&
+    isValidWalletPromptParams(event.detail.params)
   );
+}
+
+/**
+ * Validates a {@link CAIP294RequestWalletEvent} params field.
+ *
+ * @param params - The parameters to validate.
+ * @returns Whether the parameters are valid.
+ */
+function isValidWalletPromptParams(params: any): params is Record<string, any> {
+  const isValidChains =
+    params.chains === undefined ||
+    (Array.isArray(params.chains) &&
+      params.chains.every((chain: any) => typeof chain === 'string'));
+
+  const isValidAuthName =
+    params.authName === undefined || typeof params.authName === 'string';
+
+  return isValidChains && isValidAuthName;
 }
 
 /**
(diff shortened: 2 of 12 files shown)


src/initializeInpageProvider.ts
```diff
@@ -70,6 +75,10 @@ export function initializeProvider({
   });
 
   if (providerInfo) {
+    // TODO: Bring up with Jiexi & Alex, if externally_connectable isn't defined, we do not want these CAIP-294 announcements to be made,
+    const typeOrId = getTypeOrId(providerInfo.rdns);
+    const extensionId = getExtensionId(typeOrId);
+    announceWallet({ extensionId, ...providerInfo });
```
ffmcgee725: @jiexi do you want to elaborate on what we discussed regarding availability of `externally_connectable` api ?

src/CAIP294.ts
```diff
@@ -0,0 +1,213 @@
...
+}
+
+/**
+ * Intended to be used by a dapp. Forwards announced wallet to the
+ * provided handler by listening for * {@link CAIP294AnnounceWalletEvent},
+ * and dispatches an {@link CAIP294RequestWalletEvent}.
+ *
+ * @param handleWallet - A function that handles an announced wallet.
+ */
+export function requestWallet<HandlerReturnType>(
```
ffmcgee725, line 193: I attempted importing and using this from the multichain-dapp side, but was getting a weird error, so followed path of least resistance and added this code on the multichain-dapp itself. Perhaps we remove this completely ?

adonesky1:
@metamaskbot build-previews

adonesky1:
@metamaskbot publish-preview

src/initializeInpageProvider.ts
jiexi: we need to know if extension was built with the Multichain API enabled over externally_connectable. I was wrong to say it purely depended on externally_connectable. So it's basically that the extension build is flask, and that the browser is chrome

src/EIP6963.ts
```diff
@@ -27,12 +29,7 @@ declare global {
  * @property icon - The icon for the wallet. MUST be data URI.
  * @property rdns - The reverse syntax domain name identifier for the wallet.
  */
-export type EIP6963ProviderInfo = {
-  uuid: string;
-  name: string;
-  icon: string;
-  rdns: string;
-};
+export type EIP6963ProviderInfo = Omit<CAIP294WalletData, 'extensionId'>;
```
jiexi: I worry that someone will inadvertently change the `EIP6963ProviderInfo` type when they modify `CAIP294WalletData` without realizing it as this organization implies that EIP-6963 provider info is a subset of CAIP-294 provider info. 

What are your thoughts on moving the shared type (that was previously in `EIP6963ProviderInfo`) into a new type in a separate `types.ts` file that is used in this file and in the CAIP294 file? OR even more easy is maybe just restoring the change above and keeping `CAIP294WalletData` as is.

src/extension-provider/createExternalExtensionProvider.ts
```diff
@@ -72,3 +72,18 @@ function getExtensionId(typeOrId: ExtensionType) {
...
   return ids[typeOrId as keyof typeof ids] ?? typeOrId;
 }
+
+/**
+ * Gets the type or id for the given build name.
+ *
+ * @param rdns - The reverse syntax domain name identifier for the wallet.
+ * @returns The type or ID.
+ */
+export function getTypeOrId(rdns: string): string {
```
jiexi: thoughts on calling this `getBuildType` instead since that's what this function seems to do?

src/extension-provider/createExternalExtensionProvider.ts
```diff
@@ -72,3 +72,18 @@ function getExtensionId(typeOrId: ExtensionType) {
...
+  const rndsToIdDefinition: Record<string, string> = {
+    'io.metamask': 'stable',
+    'io.metamask.beta': 'beta',
+    'io.metamask.flask': 'flask',
+  };
+  return rndsToIdDefinition[rdns] ?? 'stable';
```
jiexi: I'm not sure that defaulting to stable makes sense. If rdns is some unexpected value, that would imply something funky is going on. Perhaps null instead?

jiexi: or undefined

src/EIP6963.ts
ffmcgee725: I thought about this as well. Naming wise, nothing indicates that `EIP6963ProviderInfo` extends `CAIP294WalletData` in this specific case.

My english might be busted, but what I take from your suggestion is something like
```ts
export type BaseWalletData = {
  uuid: string;
  name: string;
  icon: string;
  rdns: string;
}
```
and then both `EIP6963ProviderInfo` and `CAIP294WalletData` will extend from it yeah ? With `CAIP294WalletData` holding the extra prop `extensionId`

src/extension-provider/createExternalExtensionProvider.ts
ffmcgee725: Makes sense to me ✅ a bit more descriptive definitely

src/extension-provider/createExternalExtensionProvider.ts
ffmcgee725: I thought about that, the issue being that we would need to address the current impl of [getExtensionId()](https://github.com/MetaMask/providers/blob/main/src/extension-provider/createExternalExtensionProvider.ts#L55) since it won't take undefined as an argument. 

What's your suggestion here ? Since i'm still getting familiar, wasn't sure on how this function should handle an undefined arg passed in.

src/initializeInpageProvider.ts
jiexi: perhaps we can push the responsibility of resolving the extensionId on to the extension and then this can become something more like

```suggestion
    if(providerInfo.extensionId) {
      announceWallet(providerInfo);
    }
```

jiexi: If the extension determines itself unable to serve the Multichain API over externally_connectable, then it simply does not provider an extensionId to `initializeProvider`? Something like that?

src/EIP6963.ts
ffmcgee725: Or as you suggest, we could just revert `EIP6963ProviderInfo` despite the repeated properties. If a third type iteration shows up that has these same properties, then we could abstract it away. What do you think is best here ?

Usually my rule of thumb is, repeated properties I abstract as soon as I see them, but for functionality implementations I'm okay with 2 repetitions, once the 3rd shows then I abstract it away.

src/initializeInpageProvider.ts
```diff
@@ -70,6 +75,10 @@ export function initializeProvider({
...
     announceProvider({
```
jiexi: announceProvider and announceWallet are a bit confusing to me on first read. What are your thoughts on renaming these to something more descriptive?

src/CAIP294.ts
```diff
@@ -0,0 +1,213 @@
...
+
+/**
+ * Throws an error with link to CAIP-294 specifications.
+ *
+ * @param message - The message to include.
+ * @throws a friendly error with a link to CAIP-294.
+ */
+function throwErrorCAIP294(message: string) {
+  throw new Error(
+    `${message} See https://github.com/ChainAgnostic/CAIPs/blob/bc4942857a8e04593ed92f7dc66653577a1c4435/CAIPs/caip-294.md for requirements.`,
```
jiexi, line 228: good url for now, but we need to remember to point this to `main` once the actual PR is merged

src/CAIP294.ts
```diff
@@ -0,0 +1,213 @@
...
+
+  window.dispatchEvent(
+    new CustomEvent(CAIP294EventNames.Prompt, {
+      detail: {
+        id: 1,
+        jsonrpc: '2.0',
+        method: 'wallet_prompt',
+        params: {},
+      },
+    }),
```
jiexi, line 216: the standard is odd. I don't understand why we need to include json-rpc request in this event

jiexi: Ahh i see... 
```
window.postMessage({
  id: 2,
  jsonrpc: "2.0"
  method: "wallet_announce",
  params: walletData
});
```

we're not meant to dispatch an event here. we're meant to use `window.postMessage` instead. but why... I need to revisit the spec

src/CAIP294.ts
```diff
@@ -0,0 +1,213 @@
...
+          jsonrpc: '2.0',
+          method: 'wallet_announce',
+          params: walletData,
+        },
+      }),
+    );
+
+  _announceWallet();
+  window.addEventListener(
+    CAIP294EventNames.Prompt,
```
jiexi, line 174: this may need to be a listener on `message instead
```
window.addEventListener("message", (event) => {
  if (event.data.method === 'wallet_prompt') {
```

idk. this seems wrong. Let's revisit the spec together / I'll pull @adonesky1  in

src/initializeInpageProvider.ts
ffmcgee725: @jiexi so do you think we should only set `extensionId` if build is `flask` and browser is `chrome` ? 
This would be subject to change in the future though I can imagine...

@adonesky1 what is your suggested approach on how we should handle this here ?

src/initializeInpageProvider.ts
ffmcgee725: Any suggestions ? They are basically doing the same thing , for different improvement proposals. We could just slap the `CAIP294` and `EIP6963` respectively in there, but I would prefer better, more intuitive naming 🤔

src/initializeInpageProvider.ts
jiexi: right, this is my thinking, but this logic would happen on the extension inpage side

src/EIP6963.ts
jiexi: either works for me. You understood both of my suggestions. Perhaps the first one makes more sense

src/CAIP294.ts
jiexi: nevermind. I can't read

src/initializeInpageProvider.ts
```diff
@@ -99,3 +107,26 @@ export function setGlobalProvider(
...
+  providerInfo: CAIP294WalletData,
+): void {
+  const buildType = getBuildType(providerInfo.rdns);
+  if (buildType !== 'flask') {
+    return;
+  }
+
+  const browser = detect();
+  const extensionId =
+    (browser?.name as Browser) === 'firefox'
```
adonesky1: is this cast necessary?

src/initializeInpageProvider.ts
```diff
@@ -99,3 +107,26 @@ export function setGlobalProvider(
...
   window.dispatchEvent(new Event('ethereum#initialized'));
 }
+
+/**
+ * Announces caip294 wallet data according to build type and browser.
+ * For now, should only announce if build type is `flask`.
+ * `extensionId` is included if browser is NOT `firefox`.
+ *
+ * @param providerInfo - The provider info {@link BaseProviderInfo}that should be announced if set.
+ */
```
adonesky1:
```suggestion
/**
 * Announces [caip294](https://github.com/ChainAgnostic/CAIPs/blob/bc4942857a8e04593ed92f7dc66653577a1c4435/CAIPs/caip-294.md) wallet data according to build type and browser.
 * Until released to stable, only announces if build type is `flask`.
 * `extensionId` is included if browser is NOT `firefox` because it is only useable by browsers that support externally_connectable
 * 
 * @param providerInfo - The provider info {@link BaseProviderInfo}that should be announced if set.
 */
```

src/initializeInpageProvider.ts
```diff
@@ -99,3 +107,26 @@ export function setGlobalProvider(
...
+  if (buildType !== 'flask') {
+    return;
+  }
+
+  const browser = detect();
+  const extensionId =
+    (browser?.name as Browser) === 'firefox'
+      ? undefined
+      : getExtensionId(buildType);
+  announceWallet({ extensionId, ...providerInfo });
```
adonesky1:
```suggestion
const walletData = {
    ...providerInfo,
    ...(browser?.name !== 'firefox' && {
      extensionId: getExtensionId(buildType),
    }),
  };
  
  announceWallet(walletData);
```

adonesky1: cleaner ☝️ no dangling `extensionId` key set to undefined this way

src/CAIP294.ts
```diff
@@ -0,0 +1,210 @@
...
+  );
+}
+
+/**
+ * Intended to be used by a dapp. Forwards announced wallet to the
+ * provided handler by listening for * {@link CAIP294AnnounceWalletEvent},
+ * and dispatches an {@link CAIP294RequestWalletEvent}.
+ *
+ * @param handleWallet - A function that handles an announced wallet.
+ */
```
adonesky1, line 192: yea I suppose we should have this here since we have the equivalent in the 6963 file as well

src/initializeInpageProvider.ts
ffmcgee725: For some reason my LSP yesterday was considering `name` would only be 'react-native', therefore typescript was not satisfied with this comparison. I found it weird because in the file for [createExternalExtensionProvider](https://github.com/MetaMask/providers/blob/main/src/extension-provider/createExternalExtensionProvider.ts#L62) it wasn't an issue.
Thankfully this is not happening today 🧐 

fixed in [b3e4754](https://github.com/MetaMask/providers/pull/395/commits/b3e4754a23ac1ec64705b955c34f8b8d97fc84de)

src/initializeInpageProvider.ts
ffmcgee725: fixed in [b3e4754](https://github.com/MetaMask/providers/pull/395/commits/b3e4754a23ac1ec64705b955c34f8b8d97fc84de)

src/initializeInpageProvider.ts
ffmcgee725: fixed in [b3e4754](https://github.com/MetaMask/providers/pull/395/commits/b3e4754a23ac1ec64705b955c34f8b8d97fc84de)

src/initializeInpageProvider.test.ts
```diff
@@ -0,0 +1,72 @@
...
+    const mockProvider = {} as unknown as MetaMaskInpageProvider;
+    const dispatchEvent = jest.spyOn(window, 'dispatchEvent');
+    setGlobalProvider(mockProvider);
+
+    expect(dispatchEvent).toHaveBeenCalledTimes(1);
+    expect(dispatchEvent).toHaveBeenCalledWith(
+      new Event('ethereum#initialized'),
+    );
+  });
+});
```
jiexi, line 23: thank you for covering this code you didn't change

src/initializeInpageProvider.test.ts
```diff
@@ -0,0 +1,72 @@
...
+
+  it('should announce wallet without extensionId for firefox browser', async () => {
+    (getBuildType as jest.Mock).mockReturnValue('flask');
+    (mockProvider.request as jest.Mock).mockReturnValue({});
+
+    await announceCaip294WalletData(mockProvider, mockProviderInfo);
+
+    expect(getBuildType).toHaveBeenCalledWith(mockProviderInfo.rdns);
+    expect(announceWallet).toHaveBeenCalledWith(mockProviderInfo);
+  });
```
jiexi: what are your thoughts on nesting these last two scenarios in a `describe('build type is flask', () =>{...}`?

src/initializeInpageProvider.ts
```diff
@@ -99,3 +104,33 @@ export function setGlobalProvider(
   (window as Record<string, any>).ethereum = providerInstance;
   window.dispatchEvent(new Event('ethereum#initialized'));
 }
+
+/**
+ * Announces [caip294](https://github.com/ChainAgnostic/CAIPs/blob/bc4942857a8e04593ed92f7dc66653577a1c4435/CAIPs/caip-294.md) wallet data according to build type and browser.
```
jiexi:
```suggestion
 * Announces [CAIP-294](https://github.com/ChainAgnostic/CAIPs/blob/bc4942857a8e04593ed92f7dc66653577a1c4435/CAIPs/caip-294.md) wallet data according to build type and browser.
```

src/initializeInpageProvider.ts
```diff
@@ -99,3 +104,33 @@ export function setGlobalProvider(
...
+ * Until released to stable, `extensionId` is only retrieved from provider state if build type is `flask`.
```
jiexi:
```suggestion
 * Until released to stable, `extensionId` is only set in the `metamask_getProviderState` result if the build type is `flask`.
```

src/initializeInpageProvider.ts
```diff
@@ -99,3 +104,33 @@ export function setGlobalProvider(
...
+ * `extensionId` is included if browser is NOT `firefox` because it is only useable by browsers that support [externally_connectable](https://developer.chrome.com/docs/extensions/reference/manifest/externally-connectable).
```
jiexi:
```suggestion
 * `extensionId` is included if browser is chromium based because it is only useable by browsers that support [externally_connectable](https://developer.chrome.com/docs/extensions/reference/manifest/externally-connectable).
```

src/CAIP294.ts
```diff
@@ -0,0 +1,210 @@
...
+function isValidRequestWalletEvent(
+  event: unknown,
+): event is CAIP294RequestWalletEvent {
+  return (
+    event instanceof CustomEvent &&
+    event.type === CAIP294EventNames.Prompt &&
+    isObject(event.detail) &&
+    event.detail.method === 'wallet_prompt'
+  );
```
jiexi: is it okay if we add an `isValidWalletPromptParams` or something like that and use it here?

For now it should just do a shallow check on `chains` and `authName` that they are the right type if they are defined (i.e. array and each item is a string, or just a string respectively)

src/CAIP294.test.ts
```diff
@@ -0,0 +1,189 @@
...
+      const walletInfo = getWalletData();
+      expect(() => announceWallet(walletInfo)).not.toThrow();
+
+      delete walletInfo.extensionId;
+
+      expect(() => announceWallet(walletInfo)).not.toThrow();
+
+      walletInfo.extensionId = 'valid-string';
+      expect(() => announceWallet(walletInfo)).not.toThrow();
+    });
```
jiexi, line 101: should there be a "throws if the `extensionId` field is invalid" scenario?

src/initializeInpageProvider.test.ts
ffmcgee725: doing this, but would also make sense to wrap the one above these in a `build type is not flask` so all are contained in a respective equivalent `describe` section 👍🏾

src/CAIP294.test.ts
ffmcgee725: resolved in [07abf08](https://github.com/MetaMask/providers/pull/395/commits/07abf086656e8e45aa887ce7fb3a37c7f16d350d)

src/CAIP294.ts
ffmcgee725: resolved in [07abf08](https://github.com/MetaMask/providers/pull/395/commits/07abf086656e8e45aa887ce7fb3a37c7f16d350d)

src/initializeInpageProvider.ts
ffmcgee725: resolved in [07abf08](https://github.com/MetaMask/providers/pull/395/commits/07abf086656e8e45aa887ce7fb3a37c7f16d350d)

src/initializeInpageProvider.ts
ffmcgee725: resolved in [07abf08](https://github.com/MetaMask/providers/pull/395/commits/07abf086656e8e45aa887ce7fb3a37c7f16d350d)

src/initializeInpageProvider.ts
ffmcgee725: resolved in [07abf08](https://github.com/MetaMask/providers/pull/395/commits/07abf086656e8e45aa887ce7fb3a37c7f16d350d)

src/initializeInpageProvider.test.ts
ffmcgee725: resolved in [07abf08](https://github.com/MetaMask/providers/pull/395/commits/07abf086656e8e45aa887ce7fb3a37c7f16d350d)

src/initializeInpageProvider.ts
```diff
@@ -99,3 +104,33 @@ export function setGlobalProvider(
...
+export async function announceCaip294WalletData(
+  provider: MetaMaskInpageProvider,
+  providerInfo: CAIP294WalletData,
+): Promise<void> {
+  const buildType = getBuildType(providerInfo.rdns);
+  if (buildType !== 'flask') {
+    return;
+  }
+
+  const providerState = await provider.request<{ extensionId: string }>({
```
jiexi:
```suggestion
  const providerState = await provider.request<{ extensionId?: string }>({
```

jiexi (dismissed):

src/initializeInpageProvider.ts
ffmcgee725: resolved in [06e9771](https://github.com/MetaMask/providers/pull/395/commits/06e9771264e9bbfe9ccf74e0484dad41dff86a49)

jiexi (approved):

PR #399, opened 2025-01-22
feat: assign a `UUID` to all `BaseProvider` requests

cryptodev-2s:
This PR assigns a UUID to all `BaseProvider` requests.

### DIFF
--- a/src/BaseProvider.ts
+++ b/src/BaseProvider.ts
@@ -11,6 +11,7 @@ import type {
   Json,
 } from '@metamask/utils';
 import dequal from 'fast-deep-equal';
+import { v4 as uuidV4 } from 'uuid';
 
 import messages from './messages';
 import type { ConsoleLike, Maybe } from './utils';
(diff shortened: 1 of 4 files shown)


src/BaseProvider.ts
```diff
@@ -208,9 +209,11 @@ export abstract class BaseProvider extends SafeEventEmitter {
     const payload =
       params === undefined || params === null
         ? {
+            id: uuidV4(),
```
mcmire, line 212: Nit: Do we want to do this in `_rpcRequest`? I see that we already provide a default value for `jsonrpc` if it's not given so it seems like this check for `id` may belong there too.

mcmire: More importantly, can we add a test for this?

mcmire (commented):
I'm surprised that this hasn't caused a problem so far, but seems like a good idea? Curious to know what others think.

cryptodev-2s:
The PR was closed because this functionality should be managed by using an rpc middleware in the rpcMiddleware param available in providers, such as `createIdRemapMiddleware`, here: https://github.com/MetaMask/providers/blob/main/src/StreamProvider.ts#L58.

cryptodev-2s:
The PR was closed because this functionality should be managed by using an rpc middleware in the rpcMiddleware param available in providers, such as createIdRemapMiddleware, here: https://github.com/MetaMask/providers/blob/main/src/StreamProvider.ts#L58.

PR #400, opened 2025-01-23
refactor(StreamProvider): remove unnecessary stream multiplexing

### LINKED ISSUE #401: StreamProvider: remove unnecessary stream multiplexing
Eliminate redundant stream multiplexing within StreamProvider and move the multiplexing logic to the consumer side.

cryptodev-2s:
Eliminate redundant stream multiplexing within `StreamProvider` and move the multiplexing logic to the consumer side.

Changelog: See changelog changes

* Fixes: https://github.com/MetaMask/providers/issues/401

### DIFF
--- a/src/MetaMaskInpageProvider.ts
+++ b/src/MetaMaskInpageProvider.ts
@@ -86,8 +86,6 @@ export class MetaMaskInpageProvider extends AbstractStreamProvider {
    *
    * @param connectionStream - A Node.js duplex stream.
    * @param options - An options bag.
-   * @param options.jsonRpcStreamName - The name of the internal JSON-RPC stream.
-   * Default: `metamask-provider`.
    * @param options.logger - The logging API to use. Default: `console`.
    * @param options.maxEventListeners - The maximum number of event
    * listeners. Default: 100.
--- a/src/MetaMaskInpageProvider.ts
+++ b/src/MetaMaskInpageProvider.ts
@@ -97,14 +97,12 @@ export class MetaMaskInpageProvider extends AbstractStreamProvider {
   constructor(
     connectionStream: Duplex,
     {
-      jsonRpcStreamName = MetaMaskInpageProviderStreamName,
       logger = console,
       maxEventListeners = 100,
       shouldSendMetadata,
     }: MetaMaskInpageProviderOptions = {},
   ) {
     super(connectionStream, {
-      jsonRpcStreamName,
       logger,
       maxEventListeners,
       rpcMiddleware: getDefaultExternalMiddleware(logger),
--- a/src/MetaMaskInpageProvider.ts
+++ b/src/MetaMaskInpageProvider.ts
@@ -29,9 +29,7 @@ export type MetaMaskInpageProviderOptions = {
    * Whether the provider should send page metadata.
    */
   shouldSendMetadata?: boolean;
-
-  jsonRpcStreamName?: string | undefined;
-} & Partial<Omit<StreamProviderOptions, 'rpcMiddleware' | 'jsonRpcStreamName'>>;
+} & Partial<Omit<StreamProviderOptions, 'rpcMiddleware'>>;
 
 type SentWarningsState = {
   // methods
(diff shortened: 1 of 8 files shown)


src/extension-provider/createExternalExtensionProvider.ts
```diff
@@ -28,8 +29,16 @@ export function createExternalExtensionProvider(
     const metamaskPort = chrome.runtime.connect(extensionId) as Runtime.Port;
 
     const pluginStream = new PortStream(metamaskPort);
-    provider = new StreamProvider(pluginStream as unknown as Duplex, {
-      jsonRpcStreamName: MetaMaskInpageProviderStreamName,
+    const streamName = MetaMaskInpageProviderStreamName;
+    const mux = new ObjectMultiplex();
+    pipeline(pluginStream, mux, pluginStream, (error: Error | null) => {
+      let warningMsg = `MetaMask: Lost connection to "${streamName}".`;
```
Gudahtt: Nit: Maybe better to omit "MetaMask" from this error message, since this would be expected to be used by other extensions to connect _to_ MetaMask.

src/MetaMaskInpageProvider.ts
Gudahtt, line 1: `jsonRpcStreamName` is still present in the options type, it should be removed there as well

Gudahtt (dismissed):
Looks good! Just one last requested change for the inpage provider options.

It would be great to see the changelog entries before approving as well

Gudahtt (dismissed):
LGTM!

Gudahtt (dismissed):

src/initializeInpageProvider.ts
```diff
@@ -29,6 +33,10 @@ type InitializeProviderOptions = {
    * Whether the window.web3 shim should be set.
    */
   shouldShimWeb3?: boolean;
+  /**
+   * The name of the stream used to connect to the wallet.
+   */
+  jsonRpcStreamName?: string | undefined;
```
jiexi:
```suggestion
  jsonRpcStreamName?: string;
```

is the undefined needed?

src/initializeInpageProvider.ts
```diff
@@ -47,20 +55,30 @@ type InitializeProviderOptions = {
...
-    jsonRpcStreamName,
-    logger,
-    maxEventListeners,
-    shouldSendMetadata,
+  const mux = new ObjectMultiplex();
+  pipeline(connectionStream, mux, connectionStream, (error: Error | null) => {
+    let warningMsg = `Lost connection to "${jsonRpcStreamName}".`;
+    if (error?.stack) {
+      warningMsg += `\n${error.stack}`;
+    }
```
jiexi, line 71: thoughts on just ~~stringifying and~~ logging the entire error object?

src/initializeInpageProvider.ts
cryptodev-2s: I believe it shouldn't the parameter is already optional but this was existing I just moved the property out of `MetaMAskInpageProvider`
I just removed it

jiexi:
Why not move the stream multiplexing outside of this package entirely? I ask because I have a POC branch for adding Multichain API support over WindowPostMessage that requires me to instantiate yet another ObjectMultiplex and substream outside of the inpage provider instantiator. It seems to make more sense and is cleaner if we just remove this substream creation responsibility from these providers. I.e. the providers should just operate off the original Duplex Stream instance given to them and that's it

Gudahtt:
@jiexi I had the same thought. I'm not 100% sure how I feel about the idea, but agreed that it's worth considering.

However I still see this PR as a step towards that.

Extracting the steam multiplexing completely has some potential downsides, but this step does not (as far as I can tell).

jiexi (approved):

src/initializeInpageProvider.ts
cryptodev-2s: I tried to use the same pattern that was used earlier and as still in here https://github.com/MetaMask/providers/blob/main/src/StreamProvider.ts#L155

But I haven't nothing against making this change

PR #404, opened 2025-01-27
feat: allow `networkVersion` to be set to `null`. fire connection events based on new `isConnected` property value

jiexi:
* **BREAKING**: Interprets the provider's `networkVersion` value of `loading` to be null and for the `networkChanged` event to emit a null value
* **BREAKING**: Removes disconnect event/behavior when receiving the `loading` value for `networkVersion` and replaces it with a check against `isConnected` that is now expected in the `metamask_getProviderState` result and `metamask_chainChanged` event

Extension: https://github.com/MetaMask/metamask-extension/pull/29936
See: https://github.com/orgs/MetaMask/projects/146/views/1?filterQuery=label%3A%22team-wallet-api-platform%22+-status%3ABacklog&pane=issue&itemId=95374156&issue=MetaMask%7CMetaMask-planning%7C4039

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -45,10 +45,10 @@ const baseConfig = {
   // An object that configures minimum threshold enforcement for coverage results
   coverageThreshold: {
     global: {
-      branches: 63.88,
-      functions: 66.08,
-      lines: 68.35,
-      statements: 66.46,
+      branches: 67.58,
+      functions: 68.69,
+      lines: 68.42,
+      statements: 68.45,
     },
   },
 
--- a/jest.config.js
+++ b/jest.config.js
@@ -47,7 +47,7 @@ const baseConfig = {
     global: {
       branches: 63.88,
       functions: 66.08,
-      lines: 68.39,
+      lines: 68.35,
       statements: 66.46,
     },
   },
--- a/jest.config.js
+++ b/jest.config.js
@@ -45,10 +45,10 @@ const baseConfig = {
   // An object that configures minimum threshold enforcement for coverage results
   coverageThreshold: {
     global: {
-      branches: 67.58,
+      branches: 67.98,
       functions: 68.69,
-      lines: 68.42,
-      statements: 68.45,
+      lines: 68.62,
+      statements: 68.65,
     },
   },
 
--- a/jest.config.js
+++ b/jest.config.js
@@ -45,10 +45,10 @@ const baseConfig = {
   // An object that configures minimum threshold enforcement for coverage results
   coverageThreshold: {
     global: {
-      branches: 66.79,
-      functions: 68.69,
-      lines: 68.35,
-      statements: 68.38,
+      branches: 63.88,
+      functions: 66.08,
+      lines: 68.39,
+      statements: 66.46,
     },
   },
 
--- a/jest.config.js
+++ b/jest.config.js
@@ -45,10 +45,10 @@ const baseConfig = {
   // An object that configures minimum threshold enforcement for coverage results
   coverageThreshold: {
     global: {
-      branches: 67.98,
+      branches: 68.23,
       functions: 68.69,
-      lines: 68.62,
-      statements: 68.65,
+      lines: 68.68,
+      statements: 68.71,
     },
   },
 
--- a/src/MetaMaskInpageProvider.ts
+++ b/src/MetaMaskInpageProvider.ts
@@ -471,8 +471,15 @@ export class MetaMaskInpageProvider extends AbstractStreamProvider {
   } = {}) {
     super._handleChainChanged({ chainId, networkVersion, isConnected });
 
-    if (networkVersion !== this.#networkVersion) {
-      this.#networkVersion = networkVersion as string;
+    // The wallet will send a value of `loading` for `networkVersion` when it intends
+    // to communicate that this value cannot be resolved and should be intepreted as null.
+    // The wallet cannot directly send a null value for `networkVersion` because this
+    // would be a breaking change for existing dapps that use their own embedded MetaMask provider.
+    const targetNetworkVersion =
+      networkVersion === 'loading' ? null : networkVersion;
+
+    if (targetNetworkVersion !== this.#networkVersion) {
+      this.#networkVersion = targetNetworkVersion as string;
       if (this._state.initialized) {
         this.emit('networkChanged', this.#networkVersion);
       }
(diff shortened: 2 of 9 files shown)


jest.config.js
jiexi, line 1: try to bump these

src/utils.ts
```diff
@@ -104,14 +107,16 @@ export const isValidChainId = (chainId: unknown): chainId is string =>
...
- * string.
+ * string when available or null otherwise.
  *
  * @param networkVersion - The network version to validate.
  * @returns Whether the given network version is valid.
  */
 export const isValidNetworkVersion = (
   networkVersion: unknown,
-): networkVersion is string =>
-  Boolean(networkVersion) && typeof networkVersion === 'string';
```
Gudahtt, line 115: Any dapps (or browser extension) that use their own provider package will still have some code equivalent to this. i.e. they will throw an error if `networkVersion` is `null`.

Adding a new `isConnected` property is fine, that's non-breaking (at least for any sites that embed a provider in any way similar to ours; we don't blow up if an extra property is sent). But we may want to continue using `loading` rather than `null` for compatibility reasons.

jiexi: thanks for catching this. fixed here https://github.com/MetaMask/providers/pull/404/commits/3f24e8bacd2b96066982f47aa135259a2d22e8ff

src/StreamProvider.test.ts
```diff
@@ -399,7 +429,7 @@ describe('StreamProvider', () => {
           mockStream.notify(mockStreamName, {
             jsonrpc: '2.0',
             method: 'metamask_chainChanged',
-            params: { chainId: '0x1', networkVersion: '0x1' },
+            params: { chainId: '0x1', networkVersion: '0' },
```
adonesky1, line 432: 🤔 not following this change... Doesn't look like we're making assertions about this value. And from what I can tell this is a valid `networkVersion` so we're not really doing anything different here?

Should we make it `1` anyways to make it less confusing?

adonesky1: seems like '0' is a valid networkVersion?

src/StreamProvider.test.ts
```diff
@@ -443,44 +474,61 @@ describe('StreamProvider', () => {
           mockStream.notify(mockStreamName, {
             jsonrpc: '2.0',
             method: 'metamask_chainChanged',
-            // A "loading" networkVersion indicates the network is changing.
-            // Although the chainId is different, chainChanged should not be
-            // emitted in this case.
-            params: { chainId: '0x1', networkVersion: 'loading' },
+            params: { chainId: '0x1', networkVersion: '0', isConnected: false },
```
adonesky1, line 477: same question here, is the change of the networkVersion to `0` instead of `1` intentional for some purpose that I'm not detecting?
```suggestion
            params: { chainId: '0x1', networkVersion: '1, isConnected: false },
```

adonesky1: or I guess in this case it will more likely be 'loading' no? since `isConnected` is false?
```suggestion
            params: { chainId: '0x1', networkVersion: 'loading', isConnected: false },
```

adonesky1 (approved):
Just a few non-blocking questions about tests, otherwise LGTM!

src/StreamProvider.test.ts
jiexi: StreamProvider is only responsible for the `chainId` portion of the `metamask_chainChanged` event, so I keeo networkVersion the same here to better show that this is the only thing we are testing.

`networkVersion` may be some value other than `loading` when `isConnected` is false.

src/StreamProvider.test.ts
jiexi: The StreamProvider is initialized with a networkVersion value of `'0'`. The StreamProvider is only responsible for handling the `chainId` changes, and so I made this value `'0'` to more accurately show what this scenario is testing

src/MetaMaskInpageProvider.ts
```diff
@@ -450,25 +450,41 @@ export class MetaMaskInpageProvider extends AbstractStreamProvider {
...
-    super._handleChainChanged({ chainId, networkVersion });
-
-    if (this._state.isConnected && networkVersion !== this.#networkVersion) {
-      this.#networkVersion = networkVersion as string;
+    isConnected,
+  }: {
+    chainId?: string;
+    networkVersion?: string;
+    isConnected?: boolean;
+  } = {}) {
```
danjm, line 474: Can there ever be a difference between the `isConnected` property passed in here and `this._state.isConnected` that was used here previously?

danjm (approved):
This looks good to me. I believe it solves the problem as describe and I can't think of edge cases that it might unintentionally break

src/StreamProvider.ts
```diff
@@ -151,24 +151,23 @@ export abstract class AbstractStreamProvider extends BaseProvider {
...
+   * @param networkInfo.isConnected - Whether the network is available.
    */
   protected _handleChainChanged({
     chainId,
     networkVersion,
+    isConnected,
   }: {
     chainId?: string | undefined;
     networkVersion?: string | undefined;
+    isConnected?: boolean | undefined;
```
Gudahtt, line 170: When would this be `undefined`?

src/MetaMaskInpageProvider.ts
jiexi: yes, it is possible, but this method no longer relies on the `this._state.isConnected` value and will always fire if the networkVersion changes (which is what we want here)

src/StreamProvider.ts
jiexi: it shouldn't be, but there is some type gynamistics going on between StreamProvider and BaseProvider on this method

Gudahtt: Hmm. Well, not a blocker, but this does seem sub-optimal from an API design standpoint. If this is omitted, then we'll wrongly assume that the wallet is disconnected. A user of this API might mistakenly think this was intended. And it would be tricky to discover that's what happened if we did see that happen (particularly if it was limited to a specific dapp/scenario/etc.).

I was thinking that maybe this "provider state" object is meant to be baked-in someplace down the hierarchy, i.e. not assumed as part of BaseProvider. But no, it is used throughout. So it may as well be accurate to how we use it.

jiexi: Opting for this to be done separately later given this pattern has existed prior to this PR then

PR #405, opened 2025-02-03
fix: eth_accounts / accountsChanged behavior when wallet is locked

ffmcgee725:
This PR fixes [#3853](https://github.com/MetaMask/MetaMask-planning/issues/3853):

_Currently when the user locks the wallet (either manually or by timer) the wallet emits an `accountsChanged` event with an empty array to all connected dapps. This is very confusing since it cannot be immediately distinguished from permissions being revoked, resulting in dapps not handling these two cases gracefully._

_We should stop emitting the `accountsChanged` event when the wallet is locked._

This change addresses this by removing `handleUnlockStateChanged` functionality from the `BaseProvider`, as it was decided it does not make sense to announce any account related information when the extension locks/unlocks:

Extensions pull request that will be pulling this change for the fix: https://github.com/MetaMask/metamask-extension/pull/30067

Extension working with updated providers lib:

https://github.com/user-attachments/assets/3fcaaf9f-563a-4d65-9e4a-0dc56cfcb4f5

### DIFF
--- a/src/BaseProvider.ts
+++ b/src/BaseProvider.ts
@@ -251,7 +251,7 @@ export abstract class BaseProvider extends SafeEventEmitter {
     }
 
     if (initialState) {
-      const { accounts, chainId, isUnlocked, networkVersion } = initialState;
+      const { accounts, chainId, networkVersion } = initialState;
 
       // EIP-1193 connect
       this._handleConnect(chainId);
--- a/src/BaseProvider.ts
+++ b/src/BaseProvider.ts
@@ -129,7 +129,6 @@ export abstract class BaseProvider extends SafeEventEmitter {
     this._handleConnect = this._handleConnect.bind(this);
     this._handleChainChanged = this._handleChainChanged.bind(this);
     this._handleDisconnect = this._handleDisconnect.bind(this);
-    this._handleUnlockStateChanged = this._handleUnlockStateChanged.bind(this);
     this._rpcRequest = this._rpcRequest.bind(this);
     this.request = this.request.bind(this);
 
@@ -257,7 +256,6 @@ export abstract class BaseProvider extends SafeEventEmitter {
       // EIP-1193 connect
       this._handleConnect(chainId);
       this._handleChainChanged({ chainId, networkVersion });
-      this._handleUnlockStateChanged({ accounts, isUnlocked });
       this._handleAccountsChanged(accounts);
     }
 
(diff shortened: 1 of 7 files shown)


src/BaseProvider.ts
```diff
@@ -252,12 +251,11 @@ export abstract class BaseProvider extends SafeEventEmitter {
     }
 
     if (initialState) {
-      const { accounts, chainId, isUnlocked, networkVersion } = initialState;
+      const { accounts, chainId, networkVersion } = initialState;
 
       // EIP-1193 connect
       this._handleConnect(chainId);
       this._handleChainChanged({ chainId, networkVersion });
-      this._handleUnlockStateChanged({ accounts, isUnlocked });
```
jiexi, line 263: i'm not confident we initialize the value of  `isUnlocked` to true without this

ffmcgee725: Meaning we should still get `isUnlocked` from deconstructing initial state
```js
      const { accounts, chainId, isUnlocked, networkVersion } = initialState;
```

and set it in this function 
```js
this._state.isUnlocked = isUnlocked;
```

Or just default to `true` ?

ffmcgee725: Seems it's not necessary to set this after [df0acbf](https://github.com/MetaMask/providers/pull/405/commits/df0acbfe6d5eda2acbfdb9f093a1d91478516601)

src/MetaMaskInpageProvider.ts
```diff
@@ -409,14 +409,7 @@ export class MetaMaskInpageProvider extends AbstractStreamProvider {
...
          */
-        isUnlocked: async () => {
-          if (!this._state.initialized) {
-            await new Promise<void>((resolve) => {
-              this.on('_initialized', () => resolve());
-            });
-          }
-          return this._state.isUnlocked;
-        },
+        isUnlocked: async () => !this._state.isPermanentlyDisconnected,
```
Gudahtt, line 412: Nit: Would be nice to have a unit test for this, I don't see one. I know it wasn't covered previously (so it seems), but it looks like it would be easy to add. And coverage is decent for the rest of the file otherwise.

Gudahtt (dismissed):
One minor suggestion, but LGTM!

src/MetaMaskInpageProvider.ts
adonesky1: I'm finding it quite difficult to simulate the unrecoverable disconnection event to actually toggle the `isPermanentlyDisconnected` state... any ideas on how to achieve this?

ffmcgee725: resolved in [60bb239](https://github.com/MetaMask/providers/pull/405/commits/60bb23959907055a93c47b878d7ef8eac234bf1a)

ffmcgee725: @adonesky1 seems not much has to be done since the state `isPermanentlyDisconnected` initializes to false on `BaseProvider` parent class, so we just assert the return is it's negated value

adonesky1 (approved):
LGTM

PR #413, opened 2025-04-15
feat: Add CAIP-294 `targets` to `wallet_announce` event data. Announce CAIP-348 target in `initializeProvider`

jiexi:
* Adds `targets` to `wallet_announce` event data
* Announces a target with `caip-348` and extensionId for chromium metamask extension

See: https://github.com/pedrouid/CAIPs/pull/9
See: https://github.com/ChainAgnostic/CAIPs/pull/348
See: https://github.com/MetaMask/metamask-extension/pull/32070

### DIFF
--- a/src/CAIP294.ts
+++ b/src/CAIP294.ts
@@ -20,16 +20,24 @@ declare global {
 }
 
 /**
+ * Represents the protocol/transport supported by the wallet.
+ * @type CAIP294Target
+ * @property type - The type of the target. SHOULD reference a CAIP number in the `caip-x` format.
+ * @property value - The value specifying how to connect to the target as specified by the specification in the `type` property.
+ */
+export type CAIP294Target = { type: string; value?: unknown };
+
+/**
  * Represents the assets needed to display and identify a wallet.
  * @type CAIP294WalletData
  * @property uuid - A locally unique identifier for the wallet. MUST be a v4 UUID.
  * @property name - The name of the wallet.
  * @property icon - The icon for the wallet. MUST be data URI.
  * @property rdns - The reverse syntax domain name identifier for the wallet.
- * @property extensionId - The canonical extension ID of the wallet provider for the active browser.
+ * @property targets - The target objects specifying the protocol/transport supported by the wallet.
  */
 export type CAIP294WalletData = BaseProviderInfo & {
-  extensionId?: string | undefined;
+  targets?: CAIP294Target[];
 };
 
 /**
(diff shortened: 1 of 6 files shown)


src/initializeInpageProvider.ts
```diff
@@ -122,11 +122,20 @@ export async function announceCaip294WalletData(
   const providerState = await provider.request<{ extensionId?: string }>({
     method: 'metamask_getProviderState',
   });
+
+  const targets = [];
+
   const extensionId = providerState?.extensionId;
+  if (extensionId) {
+    targets.push({
+      type: 'caip-348',
```
jiexi, line 131: i've gone back and forth on whether this belongs in a constant, and I don't think it's necessary because we don't do this for other standards numbers. The value itself is already effectively a symbol and moving it to a constant doesn't really buy us additional value

adonesky1: Should we add a link to the CAIP PR?

src/CAIP294.ts
```diff
@@ -120,6 +128,16 @@ function isValidAnnounceWalletEvent(
...
 }
 
+/**
+ * Validates an {@link CAIP294Target} object.
+ *
+ * @param data - The {@link CAIP294Target} to validate.
+ * @returns Whether the {@link CAIP294Target} is valid.
+ */
+function isValidWalletTarget(data: unknown): data is CAIP294Target {
+  return isObject(data) && typeof data.type === 'string' && Boolean(data.type);
```
adonesky1, line 138: is `Boolean(data.type)` just to check that its not an empty string?

src/CAIP294.ts
```diff
@@ -19,17 +19,25 @@ declare global {
   }
 }
 
+/**
+ * Represents the protocol/transport supported by the wallet.
+ * @type CAIP294Target
+ * @property type - The type of the target. SHOULD reference a CAIP number in the `caip-x` format.
+ * @property value - The value specifying how to connect to the target as specified by the specification in the `type` property.
+ */
+export type CAIP294Target = { type: string; value?: unknown };
```
adonesky1, line 28: Also should we add some links to CAIP-294 here?

src/CAIP294.ts
jiexi: yeah.... it's what Erik had before i think, so i just kept the pattern

src/CAIP294.ts
jiexi: to the PR? it hasn't been merged yet so the link isn't going to be great

adonesky1: meh ok

adonesky1 (dismissed):

src/CAIP294.ts
```diff
@@ -120,6 +128,16 @@ function isValidAnnounceWalletEvent(
   );
 }
 
+/**
+ * Validates an {@link CAIP294Target} object.
```
ffmcgee725:
```suggestion
 * Validates a {@link CAIP294Target} object.
```

nit ☝🏾 correct me if I'm wrong, but this seems like the proper way to "english" ?

ffmcgee725 (approved):

PR #419, opened 2025-04-28
Refactor caip-294 announcement logic

adonesky1:
This PR refactors how we determine whether to announce CAIP-294 wallet data:

- Removes dependency on wallet rdns to determine announcement eligibility
- Adds a new `shouldAnnounceCaip294` flag to `initializeProvider` options
- Allows clients to directly control announcement behavior rather than relying on build type detection
- Removes the `getBuildType` function which is no longer needed

Instead of relying on the wallet's rdns to determine whether CAIP-294 announcements should be made (previously only "flask" builds would announce), this change gives direct control to the client. This is a more flexible approach that decouples announcement logic from wallet identification details, allowing each client to make this determination based on their own requirements.


jest.config.js
```diff
@@ -45,10 +45,10 @@ const baseConfig = {
...
   coverageThreshold: {
     global: {
-      branches: 70.68,
-      functions: 72.32,
-      lines: 71.39,
-      statements: 71.51,
+      branches: 70.0,
+      functions: 72.07,
+      lines: 70.82,
+      statements: 70.95,
```
adonesky1: I wrestled with the tests a bit and mocking became quite frustrating. I think this is fine for now 😅

jiexi (approved):

